PR #12, opened 2018-12-05
Check for presence of index field in parameter table

dweindl:


petab/core.py
```diff
@@ -73,7 +73,11 @@ def get_parameter_df(parameter_file_name):
     """Read the provided parameter file into a `pandas.Dataframe`"""
 
     parameter_df = pd.read_csv(parameter_file_name, sep='\t')
-    parameter_df.set_index(['parameterId'])
+
+    try:
+        parameter_df.set_index(['parameterId'], inplace=True)
+    except KeyError:
+        raise KeyError('Parameter table missing mandatory field `parameterId`.')
```
yannikschaelte: maybe we can for all dataframes define a list of required fields, like
```
parameter_df_req_fields = ['parameterId', 'parameterScale',...]
measurement_df_... = ...
```
That would make testing more straightforward, and the labels could be in one place.

petab/core.py
```diff
@@ -239,25 +243,27 @@ def get_simulation_to_optimization_parameter_mapping(
...
 
-    def _apply_overrides(overrides, condition_id, observable_id):
-        """Apply parameter-overrides to mapping matrix"""
+    def _apply_overrides(overrides, condition_id, observable_id, override_type):
+        """Apply parameter-overrides to mapping matrix
+        override_type: observable / noise
+        """
         for i, override in enumerate(overrides):
-            if isinstance(override, float):
+            if isinstance(override, numbers.Number):
```
yannikschaelte: good. I had the issue multiple times that numpy floats are no ordinary floats, and 64/32 is again different. as far as I understand, numbers.Number should catch all of these.

petab/lint.py
```diff
@@ -20,20 +21,28 @@ def condition_table_is_parameter_free(condition_df):
...
             return False
     return True
 
 
 def measurement_table_has_timepoint_specific_mappings(measurement_df):
     """Are there time-point or replicate specific parameter assignments in the measurement table"""
 
-    measurement_df.loc[measurement_df.noiseParameters.apply(isinstance, args=(float,)), 'noiseParameters'] = np.nan
+    measurement_df.loc[
+        measurement_df.noiseParameters.apply(isinstance, args=(numbers.Number,)), 'noiseParameters'] = np.nan
```
yannikschaelte: I guess we just had the issue with nans here, i.e. grouping difficult when all are nans? did this work now?

tests/test_lint.py
```diff
@@ -34,6 +34,10 @@ def test_condition_table_is_parameter_free():
 
     assert lint.condition_table_is_parameter_free(condition_df) is True
 
+    condition_df.fixedParameter1 =  condition_df.fixedParameter1.values.astype(int)
+
+    assert lint.condition_table_is_parameter_free(condition_df) is True
```
yannikschaelte: just a warning: I think codecov or codacy complained about assert statements. Using the unittest.assertEquals() etc. was what we did for pypesto then.

petab/core.py
dweindl: Yes, we should and will. Here i would check just for this specific field though. Don't check for more than what is really required.

petab/lint.py
dweindl: Changed.

tests/test_lint.py
dweindl: Yes, i remember. Only using it in unittests, not in production code.

PR #22, opened 2019-01-09
Feature map

yannikschaelte:
* simplified usage by not requiring to specify model file, data files etc. separately
* simplified mapping function a bit
* mapping outputs now an array of optimization parameter ids. this way, also float constants can be handled, and it is more robust, than the previously used parameter indices

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -1,0 +1,333 @@
+import pandas as pd
+import numpy as np
+import libsbml
+import sympy as sp
+import re
+import itertools
+
+from . import lint
+
+
+class OptimizationProblem:
+
+    def __init__(self, sbml_file_name, measurement_file_name, condition_file_name, parameter_file_name):
+        self.measurement_file_name = measurement_file_name
+        self.condition_file_name = condition_file_name
+        self.parameter_file_name = parameter_file_name
+        self.sbml_file = sbml_file_name
+
+        self.condition_df = get_condition_df(condition_file_name)
+        self.parameter_df = get_parameter_df(parameter_file_name)
+        self.measurement_df = get_measurement_df(measurement_file_name)
+
+        self._load_sbml()
+
+    def _load_sbml(self):
+        """Load SBML model"""
+
+        # sbml_reader and sbml_document must be kept alive. Otherwise operations on sbml_model will segfault
+        self.sbml_reader = libsbml.SBMLReader()
+        self.sbml_document = self.sbml_reader.readSBML(self.sbml_file)
+        self.sbml_model = self.sbml_document.getModel()
+
+    def get_constant_parameters(self):
+        """Provide list of IDs of parameters which are fixed (i.e. not subject to optimization, no sensitivities w.r.t. these parameters are required)"""
+
+        return list(set(self.condition_df.columns.values.tolist()) - {'conditionId', 'conditionName'})
+
+    def get_observables(self):
+        """Returns dictionary of observables definitions
+
+        see `assignment_rules_to_dict` for details"""
+
+        return get_observables(self.sbml_model)
+
+    def get_sbml_sigmas(self):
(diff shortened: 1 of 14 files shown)


petab/core.py
```diff
@@ -4,63 +4,145 @@
...
+        table.
         """
 
-        return get_sigmas(sbml_model=self.sbml_model)
+        return get_sigmas(sbml_model=self.sbml_model, remove=remove)
+
+    def get_simulation_conditions_from_measurement_df(self):
+        return get_simulation_conditions_from_measurement_df(self.measurement_df)
+
+    def map_par_sim_to_par_opt(self):
```
dweindl: Having both `map_par_sim_to_par_opt` and `get_simulation_to_optimization_parameter_mapping` is a bit confusing.

petab/core.py
```diff
@@ -321,25 +438,52 @@ def to_float_if_float(x):
...
+    
     return len(placeholders)
 
 
 def get_dynamic_parameters_from_sbml(sbml_model):
-    """Get list of non-constant parameters in sbml model"""
-
+    """
+    Get list of non-constant parameters in sbml model.
+    TODO: This is not what Y would expect. Remove?
```
dweindl: What do you mean?

petab/core.py
```diff
@@ -321,25 +438,52 @@ def to_float_if_float(x):
...
 
 
-def get_notnull_columns(df, candidates):
-    """"Return list of df-columns in candidates which are not all null/nan
+def get_dynamic_simulation_parameters(sbml_model, parameter_df):
+    par_opt_ids = get_optimization_parameters(parameter_df)
+    sbml_pars = sbml_model.getListOfParameters()
+    par_ids = [
+        p.getId() for p in sbml_pars
+        if p.getId() in par_opt_ids
```
dweindl: Please reverse condition order. comparing a single string is cheaper than lookup in a potentially long list

petab/core.py
dweindl: Can you add test cases, or change the old ones to use this function?

dweindl: Done.

petab/core.py
```diff
@@ -4,63 +4,145 @@
...
-import numbers
 
-class OptimizationProblem:
 
-    def __init__(self, sbml_file_name, measurement_file_name, condition_file_name, parameter_file_name):
-        self.measurement_file_name = measurement_file_name
-        self.condition_file_name = condition_file_name
-        self.parameter_file_name = parameter_file_name
-        self.sbml_file = sbml_file_name
+class Manager:
```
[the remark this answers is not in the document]
dweindl: `petab.Problem`?

petab/core.py
yannikschaelte: So typically there would be parameters in the SBML model that are constant and never crop up in the parameter df? Can we then safely assume that it will be *exactly* those parameters that have the attribute ``constant``?
I implemented also a 2nd function, ``get_dynamic_simulation_parameters`` (
https://github.com/ICB-DCM/PEtab/blob/feature_map/petab/core.py#L515). Then one of those should be obsolete, right? 
I currently no-where use this method anyhow, but directly extract the parameters from the parameter df, but for automatic generation of a parameter df, as you said, having the method would be good.

dweindl: > So typically there would be parameters in the SBML model that are constant and never crop up in the parameter df?

Yes, some literature-based parameters for example.

> Can we then safely assume that it will be exactly those parameters that have the attribute constant?

Would also apply to "boundary conditions", would have to check for SBML details.

dweindl: Regarding the last part: Would keep both for now and clean up a bit later

yannikschaelte: okay. since they aren't used anyway, shouldn't be a problem.

PR #31, opened 2019-01-09
Extension of validator

dweindl:


petab/core.py
```diff
@@ -157,15 +157,10 @@ def get_optimization_to_simulation_parameter_mapping(self):
...
-
 def get_condition_df(condition_file_name):
-    """
-    Read the provided condition file into a `pandas.Dataframe`.
-    Conditions are rows, parameters are columns, conditionId is index.
-    """
+    """Read the provided condition file into a `pandas.Dataframe`"""
 
     condition_df = pd.read_csv(condition_file_name, sep='\t')
-    condition_df.set_index(['conditionId'])
```
dweindl: Why removed?

petab/core.py
```diff
@@ -177,11 +172,6 @@ def get_parameter_df(parameter_file_name):
 
     parameter_df = pd.read_csv(parameter_file_name, sep='\t')
 
-    try:
-        parameter_df.set_index(['parameterId'], inplace=True)
-    except KeyError:
-        raise KeyError('Parameter table missing mandatory field `parameterId`.')
```
dweindl: Why removed?

petab/lint.py
```diff
@@ -65,6 +93,79 @@ def condition_table_is_parameter_free(condition_df):
...
 
+def check_parameter_sheet(problem):
+    check_parameter_df(problem.parameter_df)
+    parameterId_is_string(problem.parameter_df)
+    parameterScale_is_valid(problem.parameter_df)
+    parameterBounds_are_numeric(problem.parameter_df)
+    parameterEstimate_is_boolean(problem.parameter_df)
+    parameterId_is_unique(problem.parameter_df)
+    check_parameterBounds(problem.parameter_df)
+    return True
```
dweindl: Remove meaningless return value?

petab/lint.py
```diff
@@ -65,6 +93,79 @@ def condition_table_is_parameter_free(condition_df):
...
+    """Check if all entries in the parameterId column of the parameter table are string and not empty"""
+
+    for parameterId in parameter_df['parameterId']:
+        if isinstance(parameterId, str):
+            if parameterId[0].isdigit():
+                raise ValueError('parameterId ' + parameterId + ' starts with integer')
+        else:
+            raise ValueError('Empty parameterId found')
+
+    return True
```
dweindl: I would expect `parameterId_is_string` to return false if not all parameterIds are string (instead of raising)

dweindl: Same for the functions below. If called `*is*`, they should return true or false, and raise only in case of error. Otherwise please call them `assert_*` or similar.

petab/lint.py
```diff
@@ -65,6 +93,79 @@ def condition_table_is_parameter_free(condition_df):
...
+        if parameterScale not in ['lin', 'log', 'log10']:
+            raise AssertionError('Expected "lin", "log" or "log10" but got "'+parameterScale+'"')
+
+    return True
+
+def parameterBounds_are_numeric(parameter_df):
+    """Check if all entries in the lowerBound and upperBound column of the parameter table are
+    numeric """
+
+    for lowBound in parameter_df['lowerBound']:
```
dweindl: Pandas will automatically provide a float-type column if all values match the format. So instead of looping across the whole list, just check `Series.dtype`

petab/core.py
yannikschaelte: Could the index have caused problems, bc then it does not appear as a column any more, s.t. sometimes one has to do reset_index() before some query?

petab/lint.py
```diff
@@ -65,6 +93,79 @@ def condition_table_is_parameter_free(condition_df):
...
+        if not parameter_df['lowerBound'][element] <= parameter_df['upperBound'][element]:
+            raise AssertionError(f'lowerbound larger than upperBound in {parameter_df["parameterId"][element]}')
+
+
+def parameterEstimate_is_boolean(parameter_df):
+    """Check if all entries in the estimate column of the parameter table are boolean """
+
+    for estimate in parameter_df['estimate']:
+        if not int(estimate) in [True, False]:
+            raise AssertionError(f'Expected boolean but got {estimate, type(estimate)} in estimate')
```
yannikschaelte: We agreed to have estimate as 1/0, right? Then "Expected boolean" might be confusing. If we agreed to have estimate as "True/False", I would explicitly write that in the error message, i.e. "Expected boolean True/False".

petab/core.py
elbaraim: we removed the set_index because the name of that column did not appear (as Yannik mentioned)

petab/core.py
elbaraim: we removed the set_index because the name of that column did not appear (as Yannik mentioned)

petab/lint.py
elbaraim: What would you recommend?
if we only return False maybe this is not very informative...

dweindl: call it `assert_parameter_name_is_valid()`?

dweindl: (And no return value should be checked for in the caller, obviously)

PR #32, opened 2019-01-09
[WIP] Data visualization functions

dweindl:


DantongWang:
@LoosC I fixed the code problems, the current errors are because the data files used for testing does not exist. Maybe we can delete the 'test_*' files?

LoosC:
@dweindl what do you think?

> @LoosC I fixed the code problems, the current errors are because the data files used for testing does not exist. Maybe we can delete the 'test_*' files?

@dweindl what do you think? I think would be good to have an example. Can the problems somehow be circumvented?

dweindl:
It depends a bit on whether it is supposed to be a colorful example, or pure a test of whether things are running. Both are valuable. If it is an example, don't name it `test_`, otherwise `pytest` will claim it. It think it would be helpful to have a `test_` file which plots to file using all the functions you added. This quickly allows to test whether things (still) work as expected (see here for example: https://github.com/ICB-DCM/parPE/blob/master/python/tests/test_plotting.py). For that, I would not add these data files here, but use minimal examples. If you want the bigger data, simply download it on the fly (just the github URL to the file instead of the local filename) .

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+    '''
+    observableParameters = np.array(measurement_data.observableParameters)
+    noiseParameters = np.array(measurement_data.noiseParameters)
+    observableTransformation = np.array(
+        measurement_data.observableTransformation
+    )
+
+    # check if there occurs the name 'nan' in the observableParameter column
+    # , should not occur!
+    checkNaming = np.where(observableParameters == 'empty')
```
yannikschaelte: are you checking for 'nan' or 'empty'?

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+    if checkNaming[0].size > 0:
+        print('error')
```
yannikschaelte: ``raise AssertionError("NOOOOOOO I DON'T WANT THAT")``?

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # but noiseParameters should always has an entry
+    checkNaming = np.where(noiseParameters == 'empty')
+    if checkNaming[0].size > 0:
+        print('error')
```
yannikschaelte: same

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+    # help functions to check for numeric values and nans in arrays
+    isnumeric = np.vectorize(isinstance)
+    isnanVectorized = np.vectorize(np.isnan)
+
+    # assign 'empty' to missing observableParameters
+    tmp_observableParameters = observableParameters
+    tmp_ind = np.where(isnumeric(observableParameters, numbers.Number))
+    tmp_observableParameters[tmp_ind[0][
+        np.where(isnanVectorized(observableParameters[tmp_ind[0]]))]
+    ] = 'empty'
```
yannikschaelte: It would be good it the used measurement_data df is not modified in the function, to avoid problems with later calls on it. I think that the methods used here are not in-place, so there shouldn't be a problem. Otherwise, a measurement_df = copy.deepcopy(measurement_df) at the beginning of the file might help.

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+    tmp_noiseParameters[
+        np.where(isnumeric(noiseParameters, numbers.Number))
+    ] = 0
+
+    # add new column for experimentId
+    measurement_data = measurement_data.assign(
+        experimentId=pd.Series(np.zeros(len(measurement_data)))
+    )
+
+    ind_noExpId = np.where(measurement_data.experimentId == 0)
```
yannikschaelte: comment what this loop does?

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
+import pandas as pd
```
yannikschaelte: would this whole file also make sense in the main dir, not in visualization?

visualization/plot_measurement.py
```diff
@@ -0,0 +1,61 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def plot_measurementdata(DataFilePath, ConditionFilePath):
```
yannikschaelte: variable names should be lower case in python (https://www.python.org/dev/peps/pep-0008/)

visualization/plot_simulation.py
```diff
@@ -0,0 +1,55 @@
+import pandas as pd
```
yannikschaelte: plot_measurementdata and plot_simulationdata look really similar. could one improve code quality be merging both essentially?

visualization/test_experimentId.py
```diff
@@ -0,0 +1,13 @@
+import pandas as pd
+import generate_experimentId
+
+DataFilePath = "https://raw.githubusercontent.com/LoosC/Benchmark-Models/" \
```
yannikschaelte: lowercase, s.o.

visualization/test_plot.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+
+import plot_measurement
+import plot_simulation
+import matplotlib.pyplot as plt
+
+DataFilePath = "https://raw.githubusercontent.com/LoosC/Benchmark-Models/" \
```
yannikschaelte: lowercase, s.o.

visualization/generate_experimentId.py
dweindl: If you don't need the result, use `np.any`

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
...
+        print('error')
+
+    # but noiseParameters should always has an entry
+    checkNaming = np.where(noiseParameters == 'empty')
+    if checkNaming[0].size > 0:
+        print('error')
+
+    # help functions to check for numeric values and nans in arrays
+    isnumeric = np.vectorize(isinstance)
+    isnanVectorized = np.vectorize(np.isnan)
```
dweindl: Is np.vectorize necessary here?

visualization/generate_experimentId.py
```diff
@@ -0,0 +1,86 @@
+import pandas as pd
+import numpy as np
+import numbers
+
+
+def generate_experimentId(measurement_data):
```
dweindl: Please at a real test-case for this. I.e. a function which checks that for a well-defined input whether a well-defined output is generated.

visualization/generate_experimentId.py
DantongWang: Maybe also check if the 'expermentId' column already exists? Then I can always run this `generate_experimentId` after importing `measurment_data`.

dweindl:
Hi, what's the status here? Shall we chop that into smaller pieces?

DantongWang:
> Hi, what's the status here? Shall we chop that into smaller pieces?

A little bit complicated from my side. For some of the issues: merge the plot_measurement and plot_simulation, change test_* to example_*. I did it in the new branch "visualization_dose_response". The reason for not creating this new branch is that this needs the work of generating independent variable id from Jan. So still does not work for all models now. I think maybe it doesn't make since to change the same thing here as we will sooner or later merge that branch as well to the master branch?

dweindl:
@DantongWang So you are saying this pull request is obsolete, and you want to merge https://github.com/ICB-DCM/PEtab/tree/visualization_dose_response instead? And this depends on @JanHasenauer ?

DantongWang:
@dweindl No, I mean that we can first merge this to the master as it works now at least for kinetic data. and merge the other one when it is first completed.

dweindl:
> first merge this to the master

Okay, can you first please take care of the remaining review comments in here?

DantongWang:
> Okay, can you first please take care of the remaining review comments in here?
Sure I can do it. What I was trying to say is for the comments like merge the `plot_measurement` and `plot_simulation` together. I have done it in the other branch which will be sooner or later merged. And since the current branch also works properly, maybe no need to change that for now?

dweindl:
If you prefer to merge the other branch, please create a respective pull request and check if the comments from in here are all addressed there. No need for fixing things twice, yes.

DantongWang:
shall we merge it now?

petab/generate_experiment_id.py
```diff
@@ -0,0 +1,91 @@
...
+    '''
+    observable_parameters = np.array(measurement_data.observableParameters)
+    noise_parameters = np.array(measurement_data.noiseParameters)
+    observable_transformation = np.array(
+        measurement_data.observableTransformation)
+
+    # check if there is empty values in the observableParameter column, should
+    # not occur!
+    if np.any(observable_parameters == 'empty'):
+        raise AssertionError("NOOOOOOO! observable_parameter must have an "
```
yannikschaelte: :)

petab/generate_experiment_id.py
```diff
@@ -0,0 +1,91 @@
...
+                             "entry!")
+
+    # noiseParameters should always has an entry
+    if np.any(noise_parameters == 'empty'):
+        raise AssertionError("NOOOOOOO! noise_parameters must have an entry!")
```
yannikschaelte: :))

petab/generate_experiment_id.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # extract measurements with no assigned experimentId
+        ind_no_exp_id = np.where(measurement_data.experimentId == 0)[0]
+
+        # if it does not decrease there might be a missing value in the
+        # observableTransformation or noiseParameter
+        print(str(ind_no_exp_id.size) + ' measurements left to be assigned')
+        count = count + 1
+
+    print(str(count - 1) + ' experimentIds added.')
```
yannikschaelte: at some point it would be good to replace all ``print``s by logger.debug, logger.info, logger.warn etc. but I think that's not handled coherently in petab yet, so that should be fine at the moment (will change at some point).

PR #37, opened 2019-01-17
Fix rule scanning for observables and sigma

dweindl:
Iterate over rules and not parameters to avoid problems with parameter matching the conventions for observable/sigma names

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -562,17 +562,19 @@ def create_condition_df(parameter_ids, condition_ids=None):
         An pandas.DataFrame with empty given rows and columns and all nan
         values
     """
-    df = pd.DataFrame(
-        data={
-            'conditionId': condition_ids,
-        })
+
+    data = {'conditionId': []}
+    for p in parameter_ids:
+        data[p] = []
+
+    df = pd.DataFrame(data)
     df.set_index(['conditionId'], inplace=True)
 
     if not condition_ids:
         return df
 
-    for f in parameter_ids:
-        df[f] = np.nan
+    for c in condition_ids:
+        df[c] = np.nan
 
     return df
 
--- a/petab/core.py
+++ b/petab/core.py
@@ -220,16 +220,17 @@ def assignment_rules_to_dict(
     """
     result = {}
 
-    # iterate over all parameters
-    for p in sbml_model.getListOfParameters():
-        parameter_id = p.getId()
+    # iterate over rules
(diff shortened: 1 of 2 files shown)


petab/core.py
```diff
@@ -546,3 +550,48 @@ def get_notnull_columns(df, candidates):
...
+        'observableId': [],
+        'preequilibrationConditionId': [],
+        'simulationConditionId': [],
+        'measurement': [],
+        'time': [],
+        'observableParameters': [],
+        'noiseParameters': [],
+        'observableTransformation': [],
+        'noiseDistribution': [],
+    })
```
yannikschaelte: Just wondering: It might be good to limit the number of times all column names are defined as much as possible, to avoid mistakes (e.g. misspellings) and simplify updates of the data format.

petab/core.py
```diff
@@ -546,3 +550,48 @@ def get_notnull_columns(df, candidates):
...
+        data[p] = []
+
+    df = pd.DataFrame(data)
+    df.set_index(['conditionId'], inplace=True)
+
+    if not condition_ids:
+        return df
+
+    for c in condition_ids:
+        df[c] = np.nan
```
yannikschaelte: ah, that's what having the id as index is good for ...

petab/core.py
```diff
@@ -220,16 +220,17 @@ def assignment_rules_to_dict(
     """
     result = {}
 
-    # iterate over all parameters
-    for p in sbml_model.getListOfParameters():
-        parameter_id = p.getId()
+    # iterate over rules
+    for rule in sbml_model.getListOfRules():
+        if rule.getTypeCode() != libsbml.SBML_ASSIGNMENT_RULE:
```
yannikschaelte: what cases does this exclude e.g.? if you ran into an error without this, would be good to document

dweindl: It allows you to have parameters named `observable_*`, `sigma_*` independently of whether they are used as targets of assignment rules. It is pretty much documented in the markdown document. The code just did something different...

petab/core.py
dweindl: ... and to make lookups faster.

petab/core.py
dweindl: Not sure how we can really do that. Created variables with the column names would work in case of slightly modifying the name, but for all other updates we will have to check everything again anyways.

PR #46, opened 2019-02-07
Feature serialize

yannikschaelte:


tests/test_petab.py
```diff
@@ -5,6 +5,7 @@
 import os
 import libsbml
 import numpy as np
+import pickle
```
dweindl: Add as test requirement to setup.py?

yannikschaelte: it's a python library installed by default, but I'll add it though

PR #50, opened 2019-02-10
Hackaton100219

### LINKED ISSUE #13: Ignore empty space after a name
For example:
in the measurement data file I have the name 'IR1_P ' and in the SBML model I have 'IR1_P'. It would be good to be able to automatically ignore spaces at the end of the string.

- to avoid:
Unknown observables in measurement file: {'observable_IR1_P '}

elbaraim:
New features:
- Checking presence of empty spaces at end of column names and column entry names - close #13 
- Check if all files exist when running `petablint.py`. Only if all files exist, the data format checking will be executed. TO DO: execute individual file checking.


petab/lint.py
```diff
@@ -14,6 +15,17 @@ def _check_df(df, req_cols, name):
         raise AssertionError(
             f"Dataframe {name} requires the columns {missing_cols}.")
 
+def assert_presence_of_empty_spaces(names_list, name):
```
dweindl:
```suggestion
def assert_no_trailing_whitespace(names_list, name):
```

petab/lint.py
```diff
@@ -14,6 +15,17 @@ def _check_df(df, req_cols, name):
...
+    r = re.compile("\w+\s$")
+    names_with_empty_string = list(filter(r.match, names_list))
+
+    if len(names_with_empty_string) > 0:
+        if name in ["condition", "parameter", "measurement"]:
+            raise AssertionError(
+                f"Empty spaces found at end of string in column names of {name} file.")
```
dweindl:
```suggestion
                f"Trailing whitespace in column names of {name} file.")
```

petab/lint.py
```diff
@@ -24,13 +36,24 @@ def check_condition_df(df):
...
 
 def check_measurement_df(df):
     req_cols = [
         "observableId", "preequilibrationConditionId", "simulationConditionId",
         "measurement", "time", "observableParameters", "noiseParameters",
         "observableTransformation"
     ]
+
+    for column_name in req_cols:
+        if df[column_name].dtype != np.dtype(np.float) and df[column_name].dtype != np.dtype(np.int):
```
dweindl: Check for common numeric supertype using numpy or Numbers.

petab/lint.py
```diff
@@ -46,10 +69,17 @@ def check_parameter_df(df):
             f"Parameter table has wrong index {df.index.name}."
             "expected 'parameterId'.")
 
+    assert_presence_of_empty_spaces(df.index.values, "parameterId")
```
dweindl: Isn't that already included in the following loop?

elbaraim: parameterId is not included in the `req_cols` list

petab/lint.py
dweindl: Please add a test case for this function

petab/lint.py
elbaraim: I used the following now:
`np.issubdtype(df[column_name].dtype, np.number)`

dweindl: Yeah, that's good

petab/lint.py
elbaraim: I just added the test in `tests/test_lint.py`
could you have a look?

PR #51, opened 2019-02-10
Error fixes

yannikschaelte:
Fixes two errors that prevented correct usage in many cases:
* enable numeric overrides in the measurement table (a test had screwed this up by setting everything to nan)
* create parameter scales table correctly, for each condition


petab/core.py
```diff
@@ -561,10 +559,10 @@ def split_parameter_replacement_list(list_string):
 
     def to_float_if_float(x):
         try:
-            return float(x)
+            ret = float(x)
         except ValueError:
-            return x
-
+            ret = x
+        return ret
```
dweindl: What was the problem? Was cleaner before...

yannikschaelte: oh yes, I had changed that since I assumed it might cause problems. will undo.

PR #57, opened 2019-02-13
Fix parameter mapping

yannikschaelte:
* handle non-mapped parameters due to missing data: just set all corresponding parameter mappings to nan, since these will not have an impact on the parameter estimation anyway (because there are not data ...)
* fix problem with the generation of simulation conditions: use get_simulation_conditions function that searches simulationConditionIds and preequilibrationConditionIds, and returns then a dataframe of the conditions.

Note
------

The order of simulation conditions must be the same in petab and pypesto. Thus, in pypesto corresponding adaptations were made to ensure the order is the same (previously it was not)

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -413,7 +413,8 @@ def get_optimization_to_simulation_parameter_mapping(
         parameter_df=None,
         sbml_model=None,
         par_opt_ids=None,
-        par_sim_ids=None):
+        par_sim_ids=None,
+        simulation_conditions=None):
     """
     Create array of mappings. The length of the array is n_conditions, each
     entry is an array of length n_par_sim, listing the optimization parameters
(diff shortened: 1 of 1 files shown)


tests/test_petab.py
```diff
@@ -270,7 +270,7 @@ def test_partial_override(self, condition_df_2_conditions):
                      'dynamicParameter2',
                      'obs1par1override',
                      'obs1par2cond1override',
-                     'observableParameter1_obs2',
+                     np.nan,
```
yannikschaelte: had to adapt the expected behavior here. not sure if that is ok, or if one would expect to have non-overrides still in there, and apply handle_missing_overrides only afterwards explicitly

petab/core.py
```diff
@@ -453,39 +456,110 @@ def get_optimization_to_simulation_parameter_mapping(
...
+    # apply filter
+    cur_measurement_df = measurement_df.loc[row_filter, :]
+
+    return cur_measurement_df
+
+
+def handle_missing_overrides(mapping_par_opt_to_par_sim, observable_ids):
+    """
+    Find all observable parameters and noise parameters that were not mapped,
+    and set their mapping to some arbitrary numeric value.
```
dweindl:
```suggestion
    and set their mapping to NaN.
```

petab/core.py
```diff
@@ -453,39 +456,110 @@ def get_optimization_to_simulation_parameter_mapping(
...
+    """
+    missed_vals = []
+    for i_condition, mapping_for_condition in \
+            enumerate(mapping_par_opt_to_par_sim):
+        for i_val, val in enumerate(mapping_for_condition):
+            for observable_id in observable_ids:
+                if re.match("(noise|observable)Parameter[0-9]+_"
```
dweindl: Could re.compile outside the loop. Not sure about the performance gain though.

tests/test_petab.py
dweindl: Yeah, that test was from back where I thought we would allow that...

petab/core.py
yannikschaelte: yes, probably better to make explicit what we do here.

tests/test_petab.py
yannikschaelte: but it is good that the test exists already that captures this case. so I don't have to write any :)

dweindl: My pleasure

PR #58, opened 2019-02-13
 Add constantSpeciesToParameters from AMICI. Start separate file for closely SBML-related functions

dweindl:


petab/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .core import *  # noqa: F403, F401
 from .lint import *  # noqa: F403, F401
+from .sbml import *  # noqa: F403, F401
```
yannikschaelte: at some point might be good to explicitly declare exportable parts, to avoid e.g. doublings of namings. but i don't see that problem yet.

petab/sbml.py
```diff
@@ -0,0 +1,96 @@
...
+    """Convert constant species in the SBML model to constant parameters
+
+    Arguments:
+
+    sbml_model: libsbml model instance
+
+    Returns:
+    species IDs that have been turned into constants
+
+    Raises:
```
yannikschaelte: doesn't reaise anything?

petab/sbml.py
```diff
@@ -0,0 +1,96 @@
...
+    if remove:
+        for parameter_id in result:
+            sbml_model.removeRuleByVariable(parameter_id)
+            sbml_model.removeParameter(parameter_id)
+
+    return result
+
+
+def constant_species_to_parameters(sbml_model):
+    """Convert constant species in the SBML model to constant parameters
```
yannikschaelte: where is this method needed? can one add that in the docstring?

petab/sbml.py
```diff
@@ -0,0 +1,96 @@
...
+
+    sbml_model: libsbml model instance
+
+    Returns:
+    species IDs that have been turned into constants
+
+    Raises:
+
+    """
+    transformable = []
```
yannikschaelte: transformable sounds like a bool. transformables?

petab/sbml.py
dweindl: Nope.

petab/sbml.py
dweindl: Nowhere inside PEtab, but it is quite useful when setting up models for PEtab, since we don't allow for specifying constant species in the condition table.

petab/__init__.py
dweindl: Yeah currently I don't see a point in not exporting everything. The API is still in flow and the future will show what is commonly used and what not.

PR #60, opened 2019-02-15
Add helper functions to set up SBML model for PEtab

dweindl:


petab/sbml.py
```diff
@@ -98,3 +98,199 @@ def constant_species_to_parameters(sbml_model):
...
+    """
+
+    if not check_units:
+        sbml_document.setConsistencyChecks(
+            libsbml.LIBSBML_CAT_UNITS_CONSISTENCY, False)
+
+    has_problems = sbml_document.checkConsistency()
+    if has_problems:
+        print_sbml_errors(sbml_document)
+        print('WARNING: Generated invalid SBML model. Check messages above.')
```
yannikschaelte: logger.warn? By default gives the same output, and is easier to write e.g. to file + screen.

I usually do

```python
import logging

logger = logging.getLogger(__name__)
```

yannikschaelte: same for all other print commands.

petab/sbml.py
```diff
@@ -98,3 +98,199 @@ def constant_species_to_parameters(sbml_model):
...
+
+    return not has_problems
+
+
+def print_sbml_errors(sbml_document: libsbml.SBMLDocument,
```
yannikschaelte: log_sbml_errors?

petab/sbml.py
```diff
@@ -98,3 +98,199 @@ def constant_species_to_parameters(sbml_model):
...
+                      minimum_severity=libsbml.LIBSBML_SEV_WARNING):
+    """Print libsbml errors"""
```
yannikschaelte: parameter descriptions?

petab/sbml.py
```diff
@@ -98,3 +98,199 @@ def constant_species_to_parameters(sbml_model):
...
+    rule.setVariable(assignee_id)
+    rule.setFormula(formula)
+
+    return rule
+
+
+def add_model_output(sbml_model: libsbml.Model,
+                     observable_id: str,
+                     formula: str,
+                     observable_name: None):
```
yannikschaelte:
```suggestion
                     observable_name: str = None):
```

petab/sbml.py
```diff
@@ -98,3 +98,199 @@ def constant_species_to_parameters(sbml_model):
...
+        sbml_model: Model to add output to
+        formula: Formula string for model output
+        observable_id: ID without 'observable_' prefix
+        observable_name: Any observable name
+    """
+
+    if observable_name is None:
+        observable_name = observable_id
+
+    prefixed_id = f'observable_{observable_id}'
```
yannikschaelte: check whether parameter / assignment rule exists in model already?

dweindl: I will leave that to the user. If libsbml will not complain, I will not complain. When modifying the model, it is generally a good idea to check the result afterwards, see respective functions added here.

PR #61, opened 2019-02-16
Allow creating Problem from pre-existing data frames

dweindl:


petab/core.py
```diff
@@ -38,11 +38,15 @@ class Problem:
     """
 
     def __init__(self,
-                 sbml_file,
-                 condition_file,
-                 measurement_file,
+                 sbml_file=None,
```
yannikschaelte: I see here two possible issues (correct me if wrong):
* The ``__getstate__`` and ``__setstate__`` methods only serialize file names, not the (possibly expensive) datasets. These methods are requried for parallelization. But I see the point of not requiring underlying files, but only dataframes.
* The double passing of _files and _dfs makes it unintuitive whether file names of dfs are used.

What I would suggest: In ``__init__``, only pass the dfs (these need to be kept then in ``__setstate__``, but that might be unavoidable). Second, introduce a static method ``from_files``, where the files are taken and dfs created. This could be similar to ``from_folder``, and the latter could also call the former directly.

dweindl: Sounds reasonable, generally aggreeing. I see that pickling dfs is expensive, on the other hand it is a priori not clear, that the files will exist on the receiver side in the same location. Furthermore the files might have changed in between. Same goes for the dfs which could have been changed since loading. Maybe the failsafe default should be pickling dataframes which is the real state and make it optional (and the users responsibility ensure same state) to pickle only file names.

yannikschaelte: Yes, totally agree, in particular the files could not exist on target system thing. So let's maybe just have the dfs in the PetabProblem class, and only use file names to construct them in some static method. For our problems, even large scale problems seem to have less than 100MB files, so that should not be a problem.
Or is there a reason why the file names should be remembered?

dweindl: In the longer run I thought about being able to load them, modify them and save them back all at once from within that class. However, I am not sure if that would be a common use case. Passing file names in that case wouldn't be too hard...
So, agreeing on taking them out.

yannikschaelte: Ah ok. Also sounds good of course, e.g. when the sbml model needs to be changed, or some missing columns values attachted to the tables. In these cases a passed file name would be required though.

Another point: The sbml model cannot be serialized, so this is only possible via the file name :( .

dweindl: Well, it can... as string... Not that big a deal...

dweindl: Done.

petab/core.py
```diff
@@ -521,9 +521,12 @@ def handle_missing_overrides(mapping_par_opt_to_par_sim, observable_ids):
         for i_condition, mapping_for_condition in \
                 enumerate(mapping_par_opt_to_par_sim):
             for i_val, val in enumerate(mapping_for_condition):
-                if isinstance(val, numbers.Number):
+                try:
```
yannikschaelte: If observable ids starting with ``(noise|observableParameter[0-9]+_ can`` safely be assumed to be override parameters, then one can also remove the observable_id iteration

dweindl: Let's assume that.

petab/core.py
```diff
@@ -38,87 +33,110 @@ class Problem:
...
-        self._load_dfs()
+        self.condition_df = condition_df
+        self.measurement_df = measurement_df
+        self.parameter_df = parameter_df
 
         self.sbml_reader = None
         self.sbml_document = None
-        self.sbml_model = None
-        self._load_sbml()
+        self.sbml_model = sbml_model
```
yannikschaelte: it said somewhere at some point that the sbml_read and _document must be kept in memory, since otherways future calls will fail (don't know the details). just to remark this here.

dweindl: Yes, it did. User-responsibility. We don't care about document and reader, only if we open it ourselves. This restriction will also be gone with the next libsbml release.

petab/core.py
```diff
@@ -38,87 +33,110 @@ class Problem:
...
-        self._load_dfs()
+        problem = Problem()
+        if condition_file:
+            problem.condition_df = get_condition_df(condition_file)
+        if measurement_file:
+            problem.measurement_df = get_measurement_df(measurement_file)
+        if parameter_file:
+            problem.parameter_df = get_parameter_df(parameter_file)
+        if sbml_file:
+            problem._load_sbml(sbml_file)
```
yannikschaelte: in the sense of object orientation, one could first create the dfs and model, and then pass all to a new Problem(args=)

petab/lint.py
```diff
@@ -39,11 +35,12 @@ def check_condition_df(df):
             f"Condition table has wrong index {df.index.name}."
             "expected 'conditionId'.")
 
-    assert_no_trailing_whitespace(df.index.values, "conditionId")
+    assert_no_leading_trailing_whitespace(df.index.values, "conditionId")
```
yannikschaelte: would actually be helpful to have a method that can fix such simple style issues and save to file.

petab/core.py
dweindl: Now you can also pass everything to Problem to not have to care about it yourself.

petab/lint.py
dweindl: Not sure of often this will be helpful. Mostly comes from inappropriate copy and pasting...

yannikschaelte: yeah, ok. nothing needing to be implemented right now anyway

PR #70, opened 2019-02-19
Fix autodeploy pypi secret

dweindl:


dweindl:
Can't let you verify it, sorry ;)

dweindl:
Nope...

PR #80, opened 2019-03-19
add costs

yannikschaelte:


petab/core.py
```diff
@@ -352,6 +359,57 @@ def get_sigmas(sbml_model, remove=False):
...
+    remove: bool, optional (default = False)
+        Whether to remove parameters identified as a cost from the
+        sbml model.
+    """
+    # lint
+    lint.assert_noise_distributions_valid(measurement_df)
+
+    # read noise distributions from measurement file
+
+    grouping_cols = [col for col in
```
dweindl: Need to check for all-NaN, otherwise grouping will fail

dweindl: use core.get_notnull_columns

petab/lint.py
```diff
@@ -220,6 +219,7 @@ def measurement_table_has_timepoint_specific_mappings(measurement_df):
     grouped_df2 = grouped_df.groupby(grouping_cols).size().reset_index()
 
     if len(grouped_df.index) != len(grouped_df2.index):
+        logger.warn(grouped_df)
```
dweindl: This should come with a message, not?

petab/lint.py
```diff
@@ -236,6 +236,29 @@ def measurement_table_has_observable_parameter_numeric_overrides(
     return False
 
 
+def assert_noise_distributions_valid(measurement_df):
+    """
+    Check whether there are not multiple noise distributions for an
+    observable, and that the names are correct.
+    """
+    df = measurement_df.copy()
```
dweindl: Should also check for valid values, not only for uniqueness?

PR #82, opened 2019-03-19
Fix parameter check constant

dweindl:


petab/lint.py
```diff
@@ -404,6 +405,8 @@ def assert_model_parameters_in_condition_or_parameter_table(
             continue
         if parameter_id.startswith('noiseParameter'):
             continue
+        if parameter.getConstant():
+            continue
```
yannikschaelte: I hope that's always set correctly then in the sbml models (don't remember the constant attribute being documented somewhere).

FFroehlich: constant parameters means that they cannot be assigned by AssignmentRules. It's a little bit of an abuse of notations.

dweindl: It is... Not sure how to handle it better. Alternatives: 1) removing that check  2) requiring ALL model parameters to be specific in the parameter table, which becomes even more redundant with the SBML file. So probably (1)

PR #85, opened 2019-03-20
Use entrypoints to install petablint as a command

cthoyt:
- Added missing requirement "colorama"
- Update README.md accordingly
- Move `/bin/petablint.py` into main module and add Python entrypoint

Now it's possible to call `petablint` like any other unix command (e.g., `grep`, `cat`, etc.). Python automatically takes care of this when pip installing.


cthoyt:
@yannikschaelte good call! I didn't realize that you had the develop/master dichotomy. Perhaps it might be helpful for new developers to make the `develop` branch the default one.

I've merged the changes from develop into this branch then pointed the merge to the develop branch. Looks like one of the changes (adding the colorama requirement) had even already been fixed!

PR #88, opened 2019-03-20
Do not check parameter bounds if parameter is not estimated

elbaraim:
If a parameter is fixed to a X value, petab checks for parameter bounds. This is not necessary and covered in this commit.


petab/lint.py
```diff
@@ -172,11 +172,12 @@ def check_parameterBounds(parameter_df):
     in the parameter table.
     """
     for element in range(len(parameter_df['lowerBound'])):
-        if not parameter_df['lowerBound'][element] \
-                <= parameter_df['upperBound'][element]:
-            raise AssertionError(
-                f"lowerbound larger than upperBound in parameterId "
-                f"{parameter_df.index[element]}.")
+        if int(parameter_df['estimate'][element]) is True:
```
dweindl: `int(*) is True` is never true. I think `1` is currently used.

Check `'estimate' in parameter_df` first

petab/lint.py
```diff
@@ -172,11 +172,12 @@ def check_parameterBounds(parameter_df):
...
+            if not parameter_df['lowerBound'][element] \
+                    <= parameter_df['upperBound'][element]:
+                raise AssertionError(
+                    f"lowerbound larger than upperBound in parameterId "
```
dweindl:
```suggestion
                    f"lowerBound greater than upperBound for parameterId "
```

petab/lint.py
elbaraim: in `check_parameter_df` it is already called first `assert_parameter_estimate_is_boolean(df)` and then `check_parameter_bounds(df)`

elbaraim: If you do int('1') is resulting in True
`>>> int('1') == True
True`

petab/lint.py
elbaraim: updated :+1:

petab/lint.py
dweindl: Yes, but `x == True` does not imply that `x is True == True`. The latter also checks for matching types.

elbaraim: I replaced "is" with "=="

yannikschaelte:
is the target really master? I guess it's for develop, right? that can be changed at the top of the page.

petab/lint.py
```diff
@@ -171,11 +171,12 @@ def check_parameter_bounds(parameter_df):
     in the parameter table.
     """
     for element in range(len(parameter_df['lowerBound'])):
-        if not parameter_df['lowerBound'][element] \
-                <= parameter_df['upperBound'][element]:
-            raise AssertionError(
-                f"lowerBound larger than upperBound for parameterId "
-                f"{parameter_df.index[element]}.")
+        if int(parameter_df['estimate'][element]) == True:
```
dweindl:
```suggestion
        if int(parameter_df['estimate'][element]):
```

Issue #90, opened 2019-04-02
Add support for non-constant parameter overrides in condition table


dweindl:
Need to adapt optimization to simulation parameter mapping (`core.get_optimization_to_simulation_parameter_mapping`). What else?

yannikschaelte:
Probably also account for scale of parameter according to ``core.get_optimization_to_simulation_scale_mapping``?

dweindl:
Good point, thanks. Will only allow overrides with same scale for a start.

PR #92, opened 2019-04-02
Allow for condition-specific overrides of dynamic parameters (Closes …

dweindl:
…#90)

- Remove duplicate functions
- Remove non-standard handling of constant parameters
- Replace fixed parameter names by nominal values during mapping
- Check override parameter scale where possible

- Does *not* implement recursive overrides. Not sure if we would want to have that.

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -434,9 +434,12 @@ def get_optimization_to_simulation_parameter_mapping(
         simulation_conditions=None,
         warn_unmapped=True):
     """
-    Create array of mappings. The length of the array is n_conditions, each
-    entry is an array of length n_par_sim, listing the optimization parameters
-    or constants to be mapped to the simulation parameters.
+    Create array of mappings from PEtab-problem to SBML parameters.
+
+    The length of the returned array is n_conditions, each entry is an array
+    of length n_par_sim, listing the optimization parameters or constants to
+    be mapped to the simulation parameters. NaN is used where no mapping
+    exists.
 
     Parameters
     ----------
@@ -468,8 +471,7 @@ def get_optimization_to_simulation_parameter_mapping(
         simulation_conditions = get_simulation_conditions(measurement_df)
 
     if par_sim_ids is None:
-        par_sim_ids = get_dynamic_simulation_parameters(sbml_model,
-                                                        parameter_df)
+        par_sim_ids = get_model_parameters(sbml_model)
 
     n_conditions = simulation_conditions.shape[0]
 
@@ -481,6 +483,10 @@ def get_optimization_to_simulation_parameter_mapping(
         name: idx for idx, name in enumerate(par_sim_ids)
     }
 
+    _apply_dynamic_parameter_overrides(mapping, condition_df, parameter_df,
+                                       par_sim_id_to_ix)
+
+    # apply output parameter overrides
     def _apply_overrides(
             overrides, condition_ix, observable_id, override_type):
         """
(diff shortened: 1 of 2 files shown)


petab/core.py
```diff
@@ -159,16 +159,15 @@ def get_optimization_parameters(self):
...
+        return get_model_parameters(self.sbml_model)
 
     def get_dynamic_parameters_from_sbml(self):
         """
         Provide list of IDS of parameters which are dynamic, i.e. not
         fixed.
         See get_dynamic_parameters_from_sbml.
         """
-        return get_dynamic_parameters_from_sbml(self.sbml_model)
+        return get_model_parameters(self.sbml_model)
```
yannikschaelte: delete whole method since doing the same as the one above?

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+
     handle_missing_overrides(mapping, warn=warn_unmapped)
     return mapping
 
 
+def _apply_dynamic_parameter_overrides(mapping,
+                                       condition_df: pd.DataFrame,
+                                       parameter_df: pd.DataFrame,
+                                       par_sim_id_to_ix):
+    """Apply dynamic parameter overrides from condition table (inplace).
```
```suggestion
    """Apply dynamic parameter overrides from condition table (in-place).
```

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+
+    Arguments:
+        mapping, par_sim_id_to_ix:
+            see get_optimization_to_simulation_parameter_mapping
+        condition_df, parameter_df: PEtab condition and parameter table
+    """
+    for overridee_id in condition_df.columns:
+        if overridee_id == 'conditionName':
+            continue
+        if condition_df[overridee_id].dtype != 'O':
```
yannikschaelte: what is that?

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+            continue
+
+        for condition_idx, overrider_id \
```
yannikschaelte: can we be sure that the condition_idx will be the same one as the one generated via get_simulation_conditions? where also preeq is considered.

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+
+def _check_dynamic_parameter_override(
+        overridee_id, overrider_id, parameter_df: pd.DataFrame):
+    """Check for valid replacement of parameter overridee_id by overrider_id.
+    Matching scales, etc."""
+
+    if 'parameterScale' not in parameter_df:
+        return  # Nothing to check
+
+    # in case both parameter are in parameter table, their scale
```
```suggestion
    # in case both parameters are in parameter table, their scale
```

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+    # must match.
+    if overridee_id in parameter_df.index \
```
yannikschaelte: Is the overridee_id not required to be in parameter_df? Where is that parameter-condition override documented?

petab/core.py
```diff
@@ -507,10 +512,90 @@ def _apply_overrides(
...
+    """
+
+    if parameter_df is None:
+        return
+    if 'estimate' not in parameter_df:
+        return
+
+    overrides = {row.name: row.nominalValue for _, row
+                 in parameter_df.iterrows() if row.estimate != 1}
+    print(overrides)
```
```suggestion
    logger.loggylog(overrides)
```

petab/core.py
```diff
@@ -701,24 +781,9 @@ def get_placeholders(formula_string, observable_id, override_type):
...
-    par_ids = [
-        p.getId() for p in sbml_pars
-        if p.getId().startswith("noiseParameter")
-        or p.getId().startswith("observableParameter")
-        or p.getId() in par_opt_ids
-    ]
-    return par_ids
+def get_model_parameters(sbml_model: libsbml.Model):
+    """Return list of SBML model parameter IDs"""
+    return [p.getId() for p in sbml_model.getListOfParameters()]
```
yannikschaelte: no getConstant() required any more? I.e. all parameters in the sbml are to be included here?

petab/core.py
dweindl: Wasn't sure which one to remove. This is one it is.

petab/core.py
dweindl: The type of mixed-type or python objects in pandas

petab/core.py
dweindl: We don't change any ordering here and don't try find the condition anywhere else. We are just walking over the array. I don't think anything can go wrong there.

petab/core.py
dweindl: > Is the overridee_id not required to be in parameter_df? 

Wasn't completely sure how to handle that. If a parameter is always overridden: does is still have to be in the parameter table? We don't require it for the output parameters, so I think we shouldn't require it here. 

> Where is that parameter-condition override documented?

The documentation has a "parameterId" in the sample table in "Condition table" section. Could be extended at some point.

petab/core.py
dweindl: Removing.

petab/core.py
dweindl: Yes. This needs to be handled downstream. The constant thing was kind of a hack we used, but had nothing to do with the SBML's intention. This might cause trouble in other tools. Therefore, I would not introduce that here.

dweindl:
> Could you test on e.g. the Boehm/Fujita model whether we still get the correct likelihood values?

Somebody's not trusting our unittests? :)

yannikschaelte:
> > Could you test on e.g. the Boehm/Fujita model whether we still get the correct likelihood values?
> 
> Somebody's not trusting our unittests? :)

never.

petab/core.py
yannikschaelte: Yes, I would then also not require the overridee id to be in the parameter table, since none of the columns makes sense really (boundaries, scale, ... all overridden)

yannikschaelte:
> > > Could you test on e.g. the Boehm/Fujita model whether we still get the correct likelihood values?
> > 
> > 
> > Somebody's not trusting our unittests? :)
> 
> never.

More seriously: In particular to check whether the get_dynamic_parameters thing works with the real-world models and pypesto/amici yet.

dweindl:
Boehm/Fujita work with parPE. PyPESTO will require some adaptations.

petab/core.py
dweindl: I would leave that in for now. This way things should be safe. To allow for non-matching parameter scale overrides, `get_optimization_to_simulation_scale_mapping ` will also have to be adapted, I think.

PR #94, opened 2019-04-04
Add support for non-constant parameter overrides in condition table f…

dweindl:
…or parameter_df creation (#94)

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -913,10 +913,8 @@ def create_parameter_df(sbml_model, condition_df, measurement_df,
                                                      'conditionName'})
     for overridee in condition_parameters:
         # non-numeric entries are parameter overrides
-        overrides = condition_df[overridee][np.invert(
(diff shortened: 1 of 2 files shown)


petab/core.py
```diff
@@ -911,6 +907,18 @@ def append_overrides(overrides):
         append_overrides(
             split_parameter_replacement_list(row.noiseParameters))
 
+    # Append parameter overrides from condition table
+    condition_parameters = list(
+        set(condition_df.columns.values.tolist()) - {'conditionId',
+                                                     'conditionName'})
+    for overridee in condition_parameters:
```
yannikschaelte: kinda hard to read. can you add some docs?

dweindl: Hm. Line 910? Which part do you mean?

yannikschaelte: in particular what this invert stuff does

PR #99, opened 2019-04-09
Fix assert_noise_distributions_valid

dweindl:
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -256,12 +256,12 @@ def assert_noise_distributions_valid(measurement_df):
     # check for valid values
 
     for trafo in df['observableTransformation']:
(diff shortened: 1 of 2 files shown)


petab/lint.py
```diff
@@ -337,7 +337,8 @@ def assert_overrides_match_parameter_count(measurement_df, observables, noise):
             if not len(replacements) == 1 \
                     or not isinstance(replacements[0], numbers.Number):
                 raise AssertionError(
-                    f'No placeholders specified in noise model for:\n{row}\n'
+                    f'No placeholders specified SBML-file noise model for: '
```
yannikschaelte: ? grammar?

PR #101, opened 2019-04-09
Allow for condition-specific dynamic parameter overrides with differe…

### LINKED ISSUE #100: Allow for condition-specific dynamic parameter overrides with different scales
Need to 
- [x] adapt scale mapping
- [x] remove checks for same scaling
- [x] for constant overrides, apply scaling during mapping

dweindl:
…nt scales (Closes #100)

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -534,35 +534,8 @@ def _apply_dynamic_parameter_overrides(mapping,
         for condition_idx, overrider_id \
                 in enumerate(condition_df[overridee_id]):
             if isinstance(overridee_id, str):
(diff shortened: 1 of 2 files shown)


petab/core.py
```diff
@@ -588,6 +561,19 @@ def fill_in_nominal_values(mapping, parameter_df: pd.DataFrame):
             if isinstance(val, str):
                 try:
                     mapping[i_condition][i_val] = overrides[val]
+                    # rescale afterwards. if there the parameter is not
+                    # overridden, the previous line raises and we save the
```
yannikschaelte: :)

PR #102, opened 2019-04-09
Fix parameter scale mapping for condition table parameters

dweindl:


dweindl:
@yannikschaelte : I was just wondering: was the plan to have the non-overriden condition-specific parameter as IDs or as value in the mapping table? This change assumes the former.

yannikschaelte:
> @yannikschaelte : I was just wondering: was the plan to have the non-overriden condition-specific parameter as IDs or as value in the mapping table? This change assumes the former.

What would a non-overriden parameter be? Even if a simulation parameter is constantly overriden by a fixed (optimization) parameter, I would still have the parameter here because it might get optimized. But I guess that's not what you meant.

dweindl:
Sorry, I meant parameter which are overridden by a numeric value. Currently these values are *not* filled in.

yannikschaelte:
So that would correspond to condition-specific constant parameters, as amici calls them. It might make sense to include them also here, but that then touches the same topic as yesterday whether we want to treat amici constant parameters and fixed (i.e. could but won't compute derivatives) parameters differently. Eventually, we will anyway need to revise how the amici simulation parameter vector is created ...
So, I would advocate including all simulation parameters in the mapping table, and then possibly checking later before amici compilation which we can/want to have as constant parameters.

dweindl:
> So, I would advocate including all simulation parameters in the mapping table, and then possibly checking later before amici compilation which we can/want to have as constant parameters.

By including you mean filling in numeric values? Or keep as is?

yannikschaelte:
essentially, we have this line: https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/objective/petab_import.py#L177, which will fail anyway now

yannikschaelte:
i mean filling in numeric values

dweindl:
Just talking about PEtab for now :)

yannikschaelte:
yes, but how we create the table is affected by what we want to extract how later

dweindl:
Rather the other way around. Okay, this needs to be addressed elsewhere then.

PR #106, opened 2019-04-25
Perform mapping for simulation and preequilibration condition indepen…

### LINKED ISSUE #105: Fix dynamic parameter overrides for preequilibration parameters
The current parameter mapping assumes that preequilibration and simulation condition have the same dynamic parameters. So the mapping probably fails in case preequilibration and simulation condition have different dynamic parameter overrides.

dweindl:
…dently (Fixes #105)

Also change to ID-based parameter mapping dicts

Cleanup, typehinting, doc


petab/core.py
```diff
@@ -149,28 +154,33 @@ def get_constant_parameters(self):
         to optimization, no sensitivities w.r.t. these parameters are
         required).
         """
+        warnings.warn("This function will be removed in future releases. ",
```
yannikschaelte: why?

petab/core.py
```diff
@@ -349,21 +372,19 @@ def get_sigmas(sbml_model, remove=False):
...
-    Returns dictiontary of cost definitions per observable, if specified.
+    Returns dictionary of cost definitions per observable, if specified.
+
     Looks through all parameters satisfying `sbml_parameter_is_cost` and
-    return dictionary of key: observableId, value: cost definition.
+    return as dictionary.
 
-    Parameters
-    ----------
+    Parameters:
```
yannikschaelte: is that in line with python docstring formatting? don't know.

petab/core.py
```diff
@@ -733,31 +516,49 @@ def get_unique_parameters(series):
...
     """
-    Get placeholder variables in noise or observable definition.
+    Get placeholder variables in noise or observable definition for the
+    given observable ID.
+
+    Arguments:
+        formula_string: observable formula (typically from SBML model)
+        observable_id: ID of current observable
+        override_type: 'observable' or 'noise', depending on whether `formula`
+            is for observable of for noise model
```
```suggestion
            is for observable or for noise model
```

petab/parameter_mapping.py
```diff
@@ -0,0 +1,368 @@
...
+ScaleMappingDict = Dict[str, str]
+ScaleMappingDictTuple = Tuple[ScaleMappingDict, ScaleMappingDict]
+
+
+def get_optimization_to_simulation_parameter_mapping(
+        condition_df: pd.DataFrame,
+        measurement_df: pd.DataFrame,
+        parameter_df: pd.DataFrame = None,
+        sbml_model: libsbml.Model = None,
+        simulation_conditions=None,
```
yannikschaelte:
```suggestion
        simulation_conditions = None,
```

petab/parameter_mapping.py
```diff
@@ -0,0 +1,368 @@
...
+        warn_unmapped: bool = True) -> List[ParMappingDictTuple]:
+    """
+    Create list of mappings dicts from PEtab-problem to SBML parameters.
```
yannikschaelte:
```suggestion
    Create list of mapping dicts from PEtab-problem to SBML parameters.
```

petab/parameter_mapping.py
```diff
@@ -0,0 +1,368 @@
...
+    ----------
+    condition_id: Condition ID for which to perform mapping
+
+    is_preeq: If true, output parameters will not be mapped
+
+    cur_measurement_df: Measurement sub-table for current condition
+
+    condition_df, parameter_df:
+        The dataframes in the PEtab format.
+
```
yannikschaelte:
```suggestion
```

petab/core.py
dweindl: To stop abusing sbml parameter `constant` attributes. We started using that for AMICI model generation. For that it needs to be decided elsewhere what should be considered a constant/fixed parameter. If we want to keep something similar here, it should return all parameters with estimate=False and those which are overridden with such a parameter for all conditions. But in the current form, this function is obsolete in my opinion.

petab/core.py
dweindl: In AMICI we are using the colon-version. Both are valid. I prefer the latter since it means less typing.

petab/parameter_mapping.py
dweindl: Without typehints there shouldn't be any space afaik (pep8)

dweindl:
> Hard to see what actually changed, because git cannot grasp the moving of content to separate files.

Pretty much every other line changed in the moved code.

PR #109, opened 2019-05-09
Visualization

erikadudki:


visu_withVisuSpecificationFile/plotting_config.py
```diff
@@ -0,0 +1,77 @@
...
+       ----------
+
+       visualization_specification: pandas data frame contains defined data format (visualization file)
+       ax: matplotlib.Axes
+       axx: subplot axis indices
+       axy: subplot axis indices
+       conditions: Values on x-axis
+       ms: pandas data frame containing measurement data which should be plotted
+       ind_plot: boolean vector, with size: len(rows in visualization file) x 1 with 'True' entries for rows which should be plotted
+       i: current index (row number) of row which should be plotted in visualizationSpecification file
```
yannikschaelte: more descriptive name

visu_withVisuSpecificationFile/plotting_config.py
```diff
@@ -0,0 +1,77 @@
...
+                xlabel = conditions[::-1]                   # reversing
+                conditions = range(len(conditions))[::-1]   # reversing
+                ax[axx, axy].set_xticks(range(len(conditions)), xlabel)
+            elif np.all(np.diff(conditions) > 0):
+                print('monotone increasing')
+                xlabel = conditions
+                conditions = range(len(conditions))
+                ax[axx, axy].set_xticks(range(len(conditions)), xlabel)
+            else:
+                print('Error: x-conditions do not coincide, some are mon. increasing,'\
```
yannikschaelte: if it's an error, rather ``raise ValueError(msg)``?

yannikschaelte:
also tests (such as are in the tests folder already) would be nice, maybe some easy setups to check expected behavior.

visu_withVisuSpecificationFile/get_data_to_plot.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import pandas as pd
+
+def get_data_to_plot(visualization_specification, measurement_data, simulation_data, condition_ids, i, clmn_name_unique):
+
+    '''
```
dweindl: Please use triple double quotes for docstrings.

visu_withVisuSpecificationFile/get_data_to_plot.py
```diff
@@ -0,0 +1,79 @@
...
+        ind_bool_allcond = [i for i,x in enumerate(vec_bool_allcond) if x]
+        # get intersection of ind_meas and ind_bool_allcond
+        ind_intersec = np.intersect1d(ind_meas, ind_bool_allcond)
+
+        # TODO: Here not the case: So, if entries of preequCondId, time,observableParams,noiseParams, observableTransf,
+        #  noiseDistr are not the same, then this belongs not to the data of interest, which should be in one group ->
+        #  differ these data into different groups!
+        if len(ind_intersec) != len(ind_meas):
+            unique_values = np.setdiff1d(ind_meas, ind_intersec)     # find unique values in ind_meas that are not in ind_intersec
+            print(unique_values)
```
dweindl: Why is that printed? Error? -> raise. Or be more informative.

paulstapor:
Hi,
I just tried to work a little with the new visualization. I stumbled across some things and decided I'll try to make a list of things which would be good to have/maybe change:
 * We would currently have the old and the new visualization routines, which seem not to be linked. I would like to avoid having double structures. Ideally, the new routines are a superset of the old ones and hence make the old ones superfluous.
 * The new visualization routines need data, simulation and visualization file. I have an application with data, but without simulation. This way, I can't use the new simulation routines, right? --> If so, please change it such that a "None"-simulation file handles this case automatically...
 * The new routines require a `datsetId` in the measurement file. What is the `datasetId`? `ObservableId`, `ConditionId`, `ExperimentId`, I understand. `DatasetId` is new to me... Also, most of the benchmark-collection files don't have it, as far as I can see... Maybe it's hence good to not make it mandatory?

yannikschaelte:
The notebook works, the petab/visualize package structure is functional, so I would consider this fine now :+1: 

@paulstapor is currently working on some modifications (since he currently has a project where he needs the functionality), which will also integrate these visualization routines with the previous ones.

dweindl:
I would only consider that fine once CI checks pass ;)

yannikschaelte:
> I would only consider that fine once CI checks pass ;)

that's what we're still working on ;)

PR #110, opened 2019-05-09
Fix generate example

yannikschaelte:


yannikschaelte:
@erikadudki @MerktSimon Is this still a thing (you were working on this, right?)?

MerktSimon:
At least for visualization the experimentID does not play a role, because plots are grouped by plotID and for everything else there is DataSetID.

JanHasenauer:
I think the experimentID can and should be removed. It has been replaced by the more fine-grained datasetID.

paulstapor:
Okay, I'm a little confused...
Both things are actually new, right? 

As a non-developer of PEtab, I have a rough idea what `ExperimentId` does: It's groups simulations and outputs into experiments, similar to what D2D does. However, I do not understand directly, what datasetID does. could we document that somehow?

Another point is, that the benchmark examples don't have these fields. So, if we want to make them mandatory, we need to rework he PEtab files of the benchmark examples. That's doable, but we have to keep it in mind...

yannikschaelte:
as far as I understand it, datasetId has in principle the same purpose, primarily for visualizing together data that belong together. If it's not documented, i.e. the full visualizationSpecification file, then this should be done in the petab docs indeed (@MerktSimon @erikadudki).

I think nothing will be mandatory, so the files/fields should be marked as optional in the docs.

PR #122, opened 2019-08-09
Fix, test and deprecate petab.sbml.constant_species_to_parameters

dweindl:
Function failed in case of multiple occurrences of the same species.


petab/sbml.py
```diff
@@ -106,9 +111,12 @@ def constant_species_to_parameters(sbml_model: libsbml.Model) -> list:
     # Remove from reactants and products
     for reaction in sbml_model.getListOfReactions():
         for speciesId in transformables:
-            reaction.removeReactant(speciesId)
-            reaction.removeProduct(speciesId)
-            reaction.removeModifier(speciesId)
+            while reaction.removeReactant(speciesId):
```
yannikschaelte: add comment that the loop is because the functions always only remove 1? helps code maintenance later.

dweindl: done

paulstapor:
I updated your branch, Daniel. Now the checks pass.
Feel free to merge whenever you want... 🙂

dweindl:
Thanks @paulstapor

PR #128, opened 2019-08-20
[WIP] Helper function to deal with timepoint-specific parameters

### LINKED ISSUE #125: Add support for time-point specific overrides to petab.parameter_mapping
`petab.parameter_mapping` does not support time-point specific parameter overrides. To work with the current mapping system, these measurements should be split into groups by overrides and returning them as separate conditions.

dweindl:
Closes #125

@elbaraim : your turn to make it pass ;)


petab/core.py
```diff
@@ -770,3 +770,66 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
             measurement_df,
             ['observableId', 'simulationConditionId',
              'preequilibrationConditionId', 'time'])).size().values - 1)
+
+
+def flatten_timepoint_specific_output_overrides(
+        petab_problem: Problem) -> None:
```
dweindl:
```suggestion
        petab_problem: Problem) -> Problem:
```

Please also adapt the documentation for the return value, now that we are returning something.

That the current implementation of the test passes, means that we are still modifying the original object. If we keep doing that, we should not return anything. Otherwise, we should only modify a copy. Either way should be stated in the docstring.

petab/core.py
```diff
@@ -770,3 +770,66 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+        petab_problem:
+            PEtab problem to work on
+    """
+
+    df = petab_problem.measurement_df[["observableId", "preequilibrationConditionId", "simulationConditionId"]]
+    df_unique_values = df.drop_duplicates()
+    df_new = pd.DataFrame()
+    for nrow in range(len(df_unique_values.index)):
+        df = petab_problem.measurement_df.loc[(petab_problem.measurement_df['observableId'] >= df_unique_values.iloc[nrow, :].values[0])
+                                         & (petab_problem.measurement_df['preequilibrationConditionId'] <= df_unique_values.iloc[nrow, :].values[1])
```
dweindl: The `values[n]` are a bit tough to read. Can you use string-based indexing?

petab/core.py
```diff
@@ -770,3 +770,66 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+                                         & (petab_problem.measurement_df['simulationConditionId'] <= df_unique_values.iloc[nrow, :].values[2])]
+        df_observable_parameters = df[["observableParameters", "noiseParameters"]]
+        unique_scaling = df_observable_parameters["observableParameters"].unique()
+        unique_noise = df_observable_parameters["noiseParameters"].unique()
+
+        for n_noise in range(len(unique_noise)):
+            for n_scale in range(len(unique_scaling)):
+                idxs = (df["noiseParameters"].str.find(unique_noise[n_noise]) + df["observableParameters"].str.find(
+                            unique_scaling[n_scale]))
+                tmp = df["observableId"].loc[idxs == 0]
```
dweindl: I am afraid that creating new dataframes in a triple-nested loop will scale very badly. We can give it a try though.

dweindl:
*ping*

dweindl:
*ping*

dweindl:
@elbaraim : *ping*

elbaraim:
> @elbaraim : _ping_

will take care of this now

petab/core.py
elbaraim: I return the object `Problem` as changes in the sbml model and in the measurement dataframe occur.
Could you please clarify ` If we keep doing that, we should not return anything` ?

elbaraim: I included now a short explanation

dweindl: > I return the object Problem as changes in the sbml model and in the measurement dataframe occur.

Exactly, you are modifying the model and the dataframe, but not the Problem instance (the references stay the same, only the pointed-to objects are changed).

So option A: 
- Modify the original dataframe and model, and don't return anything. This makes it the user's responsibility to keep a copy if required. Returning Problem is not necessary, because is not changed as mentioned above.

Option B:
- Create a deep copy of problem, do all modification on that, and return the copy.

elbaraim: Actually options A sounds good! Thanks for the suggestion.
So I should just replace the original file with the new one?

dweindl: For option A, everything was fine, except for the return statement that should be removed.

petab/core.py
elbaraim: done

petab/core.py
elbaraim: I agree, that it may not be optimal. Can we still use this as temporary solution? Maybe we can implement an additional issue if this becomes a problem (?)

petab/core.py
```diff
@@ -900,3 +900,94 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+        petab_problem: Problem) -> None:
+    """If the PEtab problem definition has timepoint-specific
+    observableParameters or noiseParameters for the same observable, replace
+    those by replicating the respective observable.
+
+    This is a helper function for some tools which may not support such
+    timepoint-specific mappings.
+
+    Arguments:
+        petab_problem:
```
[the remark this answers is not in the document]
elbaraim: I do not understand this review point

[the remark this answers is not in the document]
yannikschaelte: one might add a type hint here (``petab_problem: Problem``)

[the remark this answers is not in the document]
dweindl: Don't, that's redundant with the function signature. Only a matter of time until those are out of sync... :)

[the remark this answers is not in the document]
yannikschaelte: valid point :) many functio docstrings still have a ``@type``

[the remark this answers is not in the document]
dweindl: Yes, that was from the days before typehints. Typehints are way more powerful and any good documentation engine should be able to handle them. I'd consider `@type` obsolete.

petab/core.py
```diff
@@ -900,3 +900,94 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
+    # changes in sbml model
+    unique_observables = df["observableId"].unique()
+    for obs in unique_observables:
+        petab_problem.sbml_model.removeRuleByVariable("observable_" + obs)
+        petab_problem.sbml_model.removeSpecies(obs)
+        petab_problem.sbml_model.removeParameter(
+            'observable_' + obs)
+
+    for n_replicate in range(len(petab_problem.measurement_df.observableId)-1):
```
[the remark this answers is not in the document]
elbaraim: good point :+1: haven't thought of this

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+            PEtab problem to work on
+    """
+
+    df = petab_problem.measurement_df[
+        ["observableId",
+         "preequilibrationConditionId",
+         "simulationConditionId"]
+    ]
+    df_unique_values = df.drop_duplicates()
+    del df  # remove variable
```
yannikschaelte: python should have intelligent enough garbage collection to see when a variable is no longer needed, so this explicit call should not be necessary? or did you notice any problems here?

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
+    # Create empty df -> to be filled with replicate-specific observables
+    df_new = pd.DataFrame()
+
+    for nrow in range(len(df_unique_values.index)):
+        df = petab_problem.measurement_df.loc[
```
yannikschaelte: add comment: what does this do? are all rows selected that have a specific set of {observableId,preeqCondId,simCondId}?

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+            (petab_problem.measurement_df['observableId'] >=
```
yannikschaelte: why are oservableIds compared using >=, not ==?

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+            & (petab_problem.measurement_df['simulationConditionId'] <=
+               df_unique_values.loc[nrow, "simulationConditionId"])
+        ]
+
+        # get list of unique observable parameters
+        unique_sc = df["observableParameters"].unique()
+        # get list of unique noise parameters
+        unique_noise = df["noiseParameters"].unique()
+
+        for n_noise in range(len(unique_noise)):
```
yannikschaelte: what does this code block do?

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
+    # Remove already existing observables from the sbml model
+    for obs in unique_observables:
+        petab_problem.sbml_model.removeRuleByVariable("observable_" + obs)
+        petab_problem.sbml_model.removeSpecies(obs)
+        petab_problem.sbml_model.removeParameter(
+            'observable_' + obs)
+
+    # Redefine with replicate-specific observables in the sbml model
+    for n_obs in petab_problem.measurement_df["observableId"].unique():
```
```suggestion
    for obs_id in petab_problem.measurement_df["observableId"].unique():
```
since the iteration is over observable ids, right?

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
+        for n_noise in range(len(unique_noise)):
+            for n_scale in range(len(unique_sc)):
+                idxs = (
+                        df["noiseParameters"].str.find(unique_noise[n_noise]) +
+                        df["observableParameters"].str.find(unique_sc[n_scale])
+                )
+                tmp = df.loc[idxs == 0, "observableId"]
+                df.loc[idxs == 0, "observableId"] = tmp+"_"+str(
+                    n_scale+n_noise+1
```
yannikschaelte: flake8 will complain here :) always spaces around operators and between variables

petab/core.py
```diff
@@ -900,3 +900,87 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
+    # Redefine with replicate-specific observables in the sbml model
+    for n_obs in petab_problem.measurement_df["observableId"].unique():
+        sbml.add_global_parameter(
+            sbml_model=petab_problem.sbml_model,
+            parameter_id='observableParameter1_' + n_obs)
+        sbml.add_global_parameter(
+            sbml_model=petab_problem.sbml_model,
+            parameter_id='noiseParameter1_' + n_obs)
+        sbml.add_model_output(
```
yannikschaelte: ah ok, here the observables are re-added.

petab/core.py
elbaraim: no problems at all, i can just remove this

petab/core.py
elbaraim: yes for observable id, but for specific "replicate" observable id
what about `replicate_id` ?

petab/core.py
elbaraim: didn't complain :wink: but yes, I can change this

petab/core.py
elbaraim: totally right, thanks!

petab/core.py
```diff
@@ -900,3 +900,99 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+    # Get unique combinations of observableId, preequilibrationConditionId
+    # and simulationConditionId
+    df_unique_values = df.drop_duplicates()
+
+    # Loop over each unique combination
+    for nrow in range(len(df_unique_values.index)):
+        df = petab_problem.measurement_df.loc[
+            (petab_problem.measurement_df['observableId'] ==
+             df_unique_values.loc[nrow, "observableId"])
+            & (petab_problem.measurement_df['preequilibrationConditionId'] <=
```
yannikschaelte: now observableId is checked with ==, but why are the others still checked with <=?

petab/core.py
```diff
@@ -900,3 +900,99 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+               df_unique_values.loc[nrow, "simulationConditionId"])
+        ]
+
+        # Get list of unique observable parameters
+        unique_sc = df["observableParameters"].unique()
+        # Get list of unique noise parameters
+        unique_noise = df["noiseParameters"].unique()
+
+        # Loop
+        for n_noise in range(len(unique_noise)):
```
yannikschaelte: usually ``n_sth`` is used for counting, ``i/j_sth`` for iterators.

petab/core.py
```diff
@@ -900,3 +900,99 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+                # in their corresponding column
+                # (full-string matches are denoted by zero)
+                idxs = (
+                        df["noiseParameters"].str.find(unique_noise[n_noise]) +
+                        df["observableParameters"].str.find(unique_sc[n_scale])
+                )
+                tmp = df.loc[idxs == 0, "observableId"]
+                # Update the observable name to replicate-specific
+                df.loc[idxs == 0, "observableId"] = tmp+"_"+str(
+                    n_scale + n_noise + 1
```
yannikschaelte: will this really be replicate-specific? a + b + 1 --> couldn't it be that you have obs parameters and 2 noise parameters, such that e.g. 1 + 0 + 1 = 0 + 1 + 1? maybe rather ``tmp + '_' + str(n_scale) + '_' + str(n_noise)``?

petab/core.py
```diff
@@ -900,3 +900,99 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+                )
+                # Append the result in a new df
+                df_new = df_new.append(df.loc[idxs == 0])
+                # Restore the observable name in the original df
+                # (for continuation of the loop)
+                df.loc[idxs == 0, "observableId"] = tmp
```
yannikschaelte: is this restoring really necessary?

petab/core.py
```diff
@@ -900,3 +900,99 @@ def measurements_have_replicates(measurement_df: pd.DataFrame) -> bool:
...
+
```
yannikschaelte: one additional thought about the loop: one might want to check whether the newly created observable ids happen to be inside the dataframe already, e.g. whether alongside ``tmp`` also sth like ``tmp_1`` already exists, because then that would mess up things if you create one of the same name ...

petab/core.py
elbaraim: actually a new replicate should have to appear, e.g., `obs1_4`. I will address this.

elbaraim: this is already covered in the test case:
`1         obs1            condition1  ...      obsParOverride2  noiseParOverride1`
`2         obs1            condition1  ...      obsParOverride2  noiseParOverride2`

same scaling, but different sigma

petab/core.py
elbaraim: Good point.
What about if the newly created observable (named as, e.g., `obs1_1`) already exists in the df, rename the newly created observable, e.g., including an additional "_" in the name, until the criteria is met (= name not found in df) ?

petab/core.py
elbaraim: for the moment I would keep it, in case I have missed something, just to be safe

petab/core.py
yannikschaelte: yeah that would be one solution

PR #130, opened 2019-08-20
Feature prior

### LINKED ISSUE #17: Specify format for priors in parameter table

paulstapor:
Added some functionality to sample starting points according to priors, added files for a test (Fixes #17 )

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -255,6 +255,15 @@ class Problem:
                                    self.measurement_df,
                                    *args, **kwargs)
 
+    def sample_parameter_startpoints(self, n_starts: int = 100):
+        """Create starting points for optimization
+
+        See sample_parameter_startpoints
+        """
+        return sample_parameter_startpoints(self.parameter_df,
+                                            n_starts=n_starts)
+
+
 
 def get_default_condition_file_name(model_name: str, folder: str = ''):
     """Get file name according to proposed convention"""
@@ -850,8 +859,7 @@ def sample_parameter_startpoints(parameter_df: pd.DataFrame,
     # get types and parameters of priors from dataframe
     prior_list = get_priors_from_df(parameter_df)
 
-    startpoints = [sample_from_prior(prior, n_starts)
-                   for prior in prior_list]
+    startpoints = [sample_from_prior(prior, n_starts) for prior in prior_list]
 
     return np.array(startpoints)
 
--- a/petab/core.py
+++ b/petab/core.py
@@ -741,6 +741,123 @@ def create_parameter_df(sbml_model: libsbml.Model,
     return df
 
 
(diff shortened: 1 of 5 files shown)


petab/core.py
```diff
@@ -741,6 +749,128 @@ def append_overrides(overrides):
...
+
+    elif p_type == 'parameterScaleUniform':
+        sp = (p_params[1] - p_params[0]) * np.random.random((n_starts,
+                                                             )) + p_params[0]
+
+    elif p_type == 'normal':
+        sp = scale(np.random.normal(loc=p_params[0], scale=p_params[1],
+                                    size=(n_starts,)))
+
+    elif p_type == 'logNormal':
```
jvanhoefer: We need to decide on a consistent parametrization of the distributions. 
I understood from our discussion (might be my misunderstanding...), that we agreed on another parametrization, but I prefer your parametrization here. What do you think @LeonardSchmiester? :)

paulstapor: I thought that was the way we agreed on it: Introducing log-normal as an exponentiated normal distribution on the linear parameter space? What should be different?

jvanhoefer: I thought that we parametrized it as it is parametrized in scipy. But as I said, I prefer your version and just wanted to make sure, that Leonard implements the same parametrization. I will adapt my code accordingly.

dweindl:
@paulstapor : Can you please include a section in the documentation on the (meaning of the) supported values for the respective columns?

paulstapor:
> @paulstapor : Can you please include a section in the documentation on the (meaning of the) supported values for the respective columns?

Indeed, that's still missing. Will do that before merging!

petab/core.py
paulstapor: Ah...
Yes, scipy does it differently...
Yes, but we do everything via numpy here, so I would also prefer not to get involved with yet another package which has redundant functionality, but uses a different definition...

paulstapor:
Okay, I updated the documentation, also about the additional columns we're actually still discussing about (initializationPriorType, initializationPriorParameters, objectivePriorType, and objectivePriorParameters). This is yet to come, but fairly simple to implement. I think/hope, it was safe to include this in the "Extensions" part of the documentation...

paulstapor:
Can we agree on that this fixes now #17 ?

dweindl:
> Can we agree on that this fixes now #17 ?

Okay for me.

paulstapor:
May/Can/Shall I merge?
Or should I first fix #131 ?

paulstapor:
> > May/Can/Shall I merge?
> > Or should I first address #131 ?
> 
> Thats a PEtab issue. Merging in pesto should be done independently of that.

Both things (this PR and the Issue) are PEtab things...
I don't see, where pypesto comes in here...

PR #133, opened 2019-08-27
Add logo

dweindl:


yannikschaelte:
can you send a screenshot? paul and me such a funny problem recently. it might depend on what fonts are installed on your system, and whether the text in the svg has been rendered or not. if not, this needs to be done first to ensure it looks the same everywhere.

MerktSimon:
image
pdf agrees with png

MerktSimon:
seems to be fine after downloading

yannikschaelte:
do PEtab_old.svg and PEtab.svg look differently for you? https://github.com/ICB-DCM/PEtab/tree/fix_logo/doc/logo

MerktSimon:
in github they look as the two in the screenshot above.
after downloading they look like this:
image

yannikschaelte:
so the not _old one looks correct now? then I suggest we have two versions: one with the text rendered, and one in the original raw format. I'll make the necessary changes.

MerktSimon:
https://github.com/ICB-DCM/PEtab/blob/fix_logo/doc/logo/PEtab.svg looks correct now.
Thank you!

PR #139, opened 2019-09-10
Fix logo

yannikschaelte:
make sure svg text is displayed the same on all systems. thus have 2 svgs: one raw (for inkscape) and one for display.


yannikschaelte:
see also #133

PR #153, opened 2019-11-29
Fixup #151: ignore conditionName column

dweindl:


dweindl:
> this is about the table header, right? might be nicer to remove that from the dataframe in the first place.

I think I'd rather ignore it where necessary. Otherwise you always need to check whether any given function expects the dataframe with or without that column.

PR #157, opened 2019-11-30
Fix(lint) correct handling of optional columns. Check before access.

dweindl:


petab/lint.py
```diff
@@ -56,17 +56,26 @@ def check_condition_df(df: pd.DataFrame, sbml_model: libsbml.Model):
...
 def check_measurement_df(df):
-    req_cols = [
-        "observableId", "preequilibrationConditionId", "simulationConditionId",
-        "measurement", "time", "observableParameters", "noiseParameters"
+    required_columns = [
+        "observableId", "simulationConditionId", "measurement", "time"
+    ]
+    optional_columns = [
+        "preequilibrationConditionId",
+        "observableParameters", "noiseParameters"
```
erikadudki: does this mean, that the optional_columns are not necessarily required in the measurement files? I guess, then we also have to consider this in the visualization routine, I guess there they are required for now...

erikadudki: shall we then open a new issue for that, or how do we want to handle this?

yannikschaelte: you mean the visualization routines assume optional columns to always exist? yes, that should then be rewritten if possible. this issue here is only about the lint, so a separate one would make sense.

dweindl: Yes to both. In the current format description we labeled them optional. Would be good to have a couple of test cases for those vis grouping functions to ensure they can handle missing columns/values

Issue #162, opened 2019-12-02
documentation: visualizing results using petab.visualize


fbergmann:
as discussed over in the [pyPESTO](https://github.com/ICB-DCM/pyPESTO/issues/200) repo, i'm trying to visualize results obtained by pyPESTO. @yannikschaelte  commented, that i should be using the `plot_data_and_simulation` function you provide with PEtab. It would be nice if that function would take just a petab problem + a data simulation data (frame) in the same format. As that would be exactly the information I have. 

but according to the information i should be able to get some result plotted using

![](https://cl.ly/0ff6e0d18b76/Image%202019-12-02%20at%204.01.50%20PM.png)

data is attached.

[data.tsv.txt](https://github.com/ICB-DCM/PEtab/files/3911975/data.tsv.txt)
[cond.tsv.txt](https://github.com/ICB-DCM/PEtab/files/3911976/cond.tsv.txt)

yannikschaelte:
it should be possible to pass both ``data_file_path`` (for the data) and ``simulation_file_path`` (for the simulations). not sure if that changes sth ...

erikadudki:
I think the issue here is, that the 'observableParameters' column in the data-file is empty. I ran into the same issue recently. 
This has to be added into the PEtab visualization functionality, that an empty 'observableParameters' column is allowed. 
I'm working on it at the moment. 
Thanks for pointing that out!

erikadudki:
@fbergmann now it should work, could you try it again?

fbergmann:
@erikadudki it took a while to check, as it did not work immidiately. When i just reran the lines from the screenshot above, i still got an empty plot with the warning: 

`UserWarning: DatasetIds would have been available, but other grouping was requested. Consider using datasetId.`

This does make it a bit harder, as most petab files come without a datasetId, and i'd have to generate it in many cases. So it would have been nice if `observable_id_list` would have worked. But i used `datasetId` and added the simulation data (here it would have been nice if petab would accepd the `measurement` column, if no `simulatedData` column would be present). And with that i got this plot: 

![](https://cl.ly/a5a21a51f973/Image%202019-12-05%20at%2010.33.41%20AM.png)

This is what i wanted, however I would change the behavior in the drawing, such that the measurement data uses a **dashed** line and the simulationData a **solid** line.

LeonardSchmiester:
[...]

PR #164, opened 2019-12-03
Visualization, allow empty column observableParameters (measurement-file)

erikadudki:
currently, if the column 'observableParameters' in the measurement file is empty, then the plotting fails (no plots are generated). 
Refers to #162.

### DIFF
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -436,12 +436,11 @@ def get_data_to_plot(vis_spec: pd.DataFrame,
 
         # check that all entries for all columns-conditions are the same:
         # check correct observable
-        # if column in m_data.observableParameters is empty
+        bool_observable = (m_data.observableParameters[ind_meas[0]] ==
+                           m_data.observableParameters)
+        # special handling, if column in m_data.observableParameters is empty
         if np.isnan(m_data.observableParameters[ind_meas[0]]):
             bool_observable = (np.isnan(m_data.observableParameters))
-        else:
-            bool_observable = (m_data.observableParameters[ind_meas[0]] ==
-                               m_data.observableParameters)
 
         # check correct observable transformation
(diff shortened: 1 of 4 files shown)


petab/visualize/helper_functions.py
```diff
@@ -438,6 +438,9 @@ def get_data_to_plot(vis_spec: pd.DataFrame,
         # check correct observable
         bool_observable = (m_data.observableParameters[ind_meas[0]] ==
                            m_data.observableParameters)
+        # special handling, if column in m_data.observableParameters is empty
+        if np.isnan(m_data.observableParameters[ind_meas[0]]):
+            bool_observable = (np.isnan(m_data.observableParameters))
```
dweindl: Superfluous parentheses

yannikschaelte: one could somehow add a test to tests/test_visualization.py just showing that the plots are not empty even when the column is empty.

dweindl: The actual plotting makes tests so slow. Better somehow check before, that there is meaningful data to plot. Otherwise agreed.

yannikschaelte: yes. not sure if it helps to boil down the plotting test cases (less data, less conditions).

paulstapor: I would be very much in favor of actually using unit tests instead of full examples. However, this should be part of another pull-request.

erikadudki: so, what is the plan now for the tests...? Shall I add a new test, or is it fine without...?

dweindl: For me it's enough to address it with #165 , which will probably change some more things here anyways.

yannikschaelte: yes yes, if you tested it "somehow", i.e. verified that for some of your models this works now, that's fine. the tests will need to be rewritten substantially, so that's too much work now.

PR #169, opened 2019-12-03
Update format documentation with respect to data and parameter scales…

### LINKED ISSUE #154: Clarify meaning of observableTransformation and parameterScale
For these two cases it's somewhat confusing whether they indicate what has been or what should be applied to measurements and parameters.

Most recent discussion was that it might be easiest to go for specifying everything on linear scale.

I.e. for the parameters table, the specified nominalValue and bounds would be transformed according to parameterScale.

Any other thoughts?

dweindl:
… (Closes #154)

* Fix formatting

* [x] Check for required changes in validator for observable transformation
* [x] Check for required changes in validator for parameter scaling
* [x] Check for measurement positivity in case of log-observables
* [x] Check for bound positivity in case of log-parameters

### DIFF
--- a/doc/documentation_data_format.md
+++ b/doc/documentation_data_format.md
@@ -2,6 +2,7 @@
 
 This document explains the PEtab data format.
 
+
 ## Purpose
 
 Providing a standardized way for specifying parameter estimation problems in
@@ -70,8 +71,12 @@ E.g.
 ```
 observable_pErk = observableParameter1_pErk + observableParameter2_pErk*pErk
 ```
-where `observableParameter1_pErk` would be an offset, and `observableParameter2_pErk` a
-scaling parameter for the observable `pErk`. The observable parameter names have the structure: `observableParameter${indexOfObservableParameter}_${observableId}` to facilitate automatic recognition. The specific values or parameters are assigned in the *measurement table*.
+where `observableParameter1_pErk` would be an offset, and
+`observableParameter2_pErk` a scaling parameter for the observable `pErk`.
+The observable parameter names have the structure:
+`observableParameter${indexOfObservableParameter}_${observableId}` to
+facilitate automatic recognition. The specific values or parameters are
+assigned in the *measurement table*.
 
 
 ### Noise model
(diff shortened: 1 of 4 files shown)


doc/documentation_data_format.md
```diff
@@ -291,16 +306,19 @@ Additional columns may be added.
...
   Upper bound of the parameter used for optimization.
   Optional, if `estimate==0`.
+  Must be provided in linear space, independent of `parameterScale`.
 
 - `nominalValue` [NUMERIC]
 
-  Some parameter value (scale as specified in `parameterScale`) to be used if
+  Some parameter value to be used if
   the parameter is not subject to estimation (see `estimate` below).
+  Must be provided in linear space, independent of `parameterScale`.
```
yannikschaelte: :+1:

doc/documentation_data_format.md
```diff
@@ -314,24 +332,27 @@ Additional columns may be added.
...
-    - parameterScaleLaplace: Laplace prior on original parameter scale
+  `objectivePriorType`, respectively.
+
+  Possible prior types are (see also Extensions):
+
+    - *uniform*: flat prior on linear parameters
+    - *normal*: Gaussian prior on linear parameters
+    - *laplace*: Laplace prior on linear parameters
+    - *logNormal*: exponentiated Gaussian prior on linear parameters
+    - *logLaplace*: exponentiated Laplace prior on linear parameters
```
yannikschaelte: why don't we use snake_case here?

doc/documentation_data_format.md
```diff
@@ -379,72 +400,74 @@ order:
...
+  The type of the corresponding plot, can be `LinePlot` or `BarPlot`. Default
+  is `LinePlot`.
+
+- `plotTypeData`
+
+  The type how replicates should be handled, can be `MeanAndSD`,
+  `MeanAndSEM`, `replicate` (for plotting all replicates separately), or
+  `provided` (if numeric values for the noise level are provided in the
+  measurement table). Default is `MeanAndSD`.
+
```
yannikschaelte: just aligning I assume.

petab/lint.py
```diff
@@ -224,15 +224,20 @@ def assert_parameter_bounds_are_numeric(parameter_df):
...
-                    <= parameter_df['upperBound'][element]:
+    for _, row in parameter_df.iterrows():
+        if int(row['estimate']):
+            if not row['lowerBound'] <= row['upperBound']:
                 raise AssertionError(
-                    f"lowerbound greater than upperBound for parameterId "
-                    f"{parameter_df.index[element]}.")
+                    f"lowerBound greater than upperBound for parameterId "
+                    f"{row.name}.")
+            if (row['lowerBound'] <= 0.0 or row['upperBound'] < 0.0) \
```
yannikschaelte: you check lower <= upper before, so here lower < 0 is enough.

doc/documentation_data_format.md
dweindl: Because we don't use it anywhere in the column headers. Meanwhile I wish we had used it there :). But the initial capital is inconsistent, agreed. Maybe to be be discussed elsewhere?

petab/lint.py
dweindl: I'd leave it as it doesn't hurt.

petab/lint.py
```diff
@@ -224,15 +224,20 @@ def assert_parameter_bounds_are_numeric(parameter_df):
...
+                    and row['parameterScale'] in ['log', 'log10']:
```
elbaraim: Then we also want to provide the parameter bounds in LINEAR space? I thought only the `nominalValues` column.

elbaraim: ?

dweindl: Yes, as mentioned in https://github.com/ICB-DCM/PEtab/issues/154#issue-530236404

elbaraim: ah yes! thanks

PR #161, opened 2019-12-01
Setup documentation for RTD

dweindl:
* Create sphinx config and basic documentation pages
* Fix docstrings for sphinx processing
* Fix markdown tables
* Setup deployment

https://petab.readthedocs.io/en/latest/
To be improved later on. Closes #135.

PR #149, opened 2019-11-28
Fix create_parameter_df: Exclude rule targets

dweindl:
Closes #148

PR #172, opened 2019-12-04
Release 0.0.0a17

### LINKED ISSUE #43: Extend validator to check if all required parameters are present in parameter table
See  #42 for what to check.

Should be added to `petab/lint.py`:
> def check_parameter_sheet(problem):

EDIT: Also check that there are no additional parameters

### LINKED ISSUE #111: Remove experimentID
as discussed in #110 , remove the experimentID in favor of datasetID. don't know if a replacement for petab.generate_experiment_id will be needed.

### LINKED ISSUE #118: Specific error messages if sbml not well-defined
Since we have some specific sbml requirements, it might be informative for the user to give explicit error messages, e.g. "you need to define sigmas" etc., as far as possible. These things are a bit difficult to detect currently.

### LINKED ISSUE #121: Handle None in petab.core.split_parameter_replacement_list
Would be convenient if this function could handle None in addition to empty strings, in case column types have not been inferred correctly.

Currently:
```
"petab/core.py", line 544, in split_parameter_replacement_list
    result = [x.strip() for x in list_string.split(delim) if len(x.strip())]
AttributeError: 'NoneType' object has no attribute 'split'
```

### LINKED ISSUE #125: Add support for time-point specific overrides to petab.parameter_mapping

### LINKED ISSUE #150: Validator: verify condition table column names occur as model parameters

dweindl:
Data format: *No changes*

Library:
* Extended visualization support
* Add helper function and test case to deal with timepoint-specific parameters
  flatten_timepoint_specific_output_overrides (#128) (Closes #125)
* Fix get_noise_distributions: so far we got 'normal' everywhere due to 
  wrong grouping (#147)
* Fix create_parameter_df: Exclude rule targets (#149)
* Verify condition table column names occur as model parameters
  (Closes #150) (#151)
* More informative error messages in case of wrongly set observable and
  noise parameters (Closes #118) (#155)
* Update doc for copasi import and github installation (#158) 
* Extend validator to check if all required parameters are present in
  parameter table (Closes #43) (#159)
* Setup documentation for RTD (#161)
* Handle None in petab.core.split_parameter_replacement_list (Closes #121) 
* Fix(lint) correct handling of optional columns. Check before access.
* Remove obsolete generate_experiment_id.py (Closes #111) #166


dweindl:
Closing in order to cleanup git history mess

PR #178, opened 2019-12-12
Feature 154 parameter scale

elbaraim:
reference to 
### Clarify meaning of observableTransformation and parameterScale #154 

- Check positivity (>=0) of all entries in `nominalValue`, `upperBound` and `lowerBound` from parameter table.
- Corresponding test created.

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -215,8 +215,8 @@ def assert_parameter_scale_is_valid(parameter_df):
 
 def assert_parameter_values_and_bounds_are_numeric(parameter_df):
     """
-    Check if all entries in the nominalValue, lowerBound and upperBound columns of the
-    parameter table are numeric.
+    Check if all entries in the nominalValue, lowerBound and
+    upperBound columns of the parameter table are numeric.
     """
     parameter_df["lowerBound"].apply(float).all()
     parameter_df["upperBound"].apply(float).all()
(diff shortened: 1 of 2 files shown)


elbaraim:
> Hmm. This seems to be redundant with https://github.com/ICB-DCM/PEtab/pull/169/files, isn't it?

I think the positivity of `nominalValues` it was not implemented (?)

dweindl:
> > Hmm. This seems to be redundant with https://github.com/ICB-DCM/PEtab/pull/169/files, isn't it?
> 
> I think the positivity of `nominalValues` it was not implemented (?)

That's true, but they only need to be positive if log-scale is selected.

elbaraim:
> > > Hmm. This seems to be redundant with https://github.com/ICB-DCM/PEtab/pull/169/files, isn't it?
> > 
> > 
> > I think the positivity of `nominalValues` it was not implemented (?)
> 
> That's true, but they only need to be positive if log-scale is selected.

Totally makes sense that they need to be positive if log-scale. However, shouldn't we still infer positivity even in linear-scale? Negative parameters do not biologically make sense... What do you think about this?

dweindl:
> Negative parameters do not biologically make sense... What do you think about this?

That is true for rate constants, volumes, etc.. However, there may well be all kinds of other parameters for which it may make sense to be negative.

FFroehlich:
> > Negative parameters do not biologically make sense... What do you think about this?
> 
> That is true for rate constants, volumes, etc.. However, there may well be all kinds of other parameters for which it may make sense to be negative.

Energies, Charges, ...

elbaraim:
True! Thank you for the feedback then I would suggest to remove this PR

PR #183, opened 2019-12-15
YAML file for grouping PEtab files

### LINKED ISSUE #49: Handling of parameter estimation problems with multiple sbml models
In this [model](https://github.com/LoosC/Benchmark-Models/tree/hackathon/hackathon_contributions_new_data_format/Becker_Science2010) there are to model files and, therefore, two measurement data files and experimental condition files. But only one for the parameter file. The naming is:
- model_Becker_Science2010__BaF3_Exp.xml
- model_Becker_Science2010__binding.xml

At the moment, this throws an error in PEtab (as it does not follow the standard naming we established).

How should we circumvent this problem?
- Should we allow for finding a list of model names in the model folder and then check the respective pairs?
- Should we re-define the naming convention for such exception cases?

dweindl:
Closes #49 

* [x] Validation of contents (e.g. files exist)
* [x] petab.Problem from yaml
* [x] Adapt schema to multiple measurement files
* [x] Use in `petablint`
* [x] Implement `petab.CompositeProblem`
* [x] Multiple measurement files for `petab.Problem` (only in from_files, merge during loading)

To be done, but out of scope here:
* [ ] Add more human-readable format description
* [ ] Visualization files?
* [ ] Implement mapping functions for `CompositeProblem`
* [ ] Implement validation functions for `CompositeProblem`
* [ ] Option to specify training and validation data in YAML?


dweindl:
@yannikschaelte : How do we want to handle problems with multiple models with `petab.Problem`? Either we change the structure there, which will require a load of downstream changes, or we introduce something like `petab.CompositeProblem`, that manages a bunch of `petab.Problem`s. Not completely sure, but I think I prefer the latter, as the default still seems to be having only one model.

yannikschaelte:
> @yannikschaelte : How do we want to handle problems with multiple models with `petab.Problem`? Either we change the structure there, which will require a load of downstream changes, or we introduce something like `petab.CompositeProblem`, that manages a bunch of `petab.Problem`s. Not completely sure, but I think I prefer the latter, as the default still seems to be having only one model.

I would agree to keep the single-anything `petab.Problem`, since that will remain the default, and have a `petab.CompositeProblem` in addition. However, it won't just hold a list of `petab.Problem`s, but need to overwrite/aggregate the single functions that the Problem offers in different ways, if we have e.g. just one parameters file.
Whether to generate a Problem or CompositeProblem could easily be automatized when reading in the yaml.

doc/documentation_data_format.md
```diff
@@ -366,9 +366,42 @@ Additional columns may be added.
...
 
-## Parameter estimation problems combining multiple models
+### Additional optional columns
 
-[**Issue #49**](https://github.com/ICB-DCM/PEtab/issues/49)
+Extra columns:
+
+- `hierarchicalOptimization` (optional)
+
+  hierarchicalOptimization: 1 if parameter is optimized using hierarchical
```
yannikschaelte: not completely sure yet if this will be applicable, or if we need more info (e.g. on whether it's an offset, scaling, sigma etc.; this info should be possible to extract from the mathml, but there may be special cases).

yannikschaelte: a pypesto implementation is in a preliminary state, for the moment I think we can just leave this here as is, since it's optional anyway.

doc/documentation_data_format.md
```diff
@@ -469,44 +502,31 @@ order:
...
-### Parameter table
+## YAML file for grouping files
 
-Extra columns:
+To link the SBML model, measurement table, condition table, etc. in an
+unambiguous way, we use a [YAML](https://yaml.org/) file.
 
-- `hierarchicalOptimization` (optional)
+This file also allows specifying a PEtab version (as the format is not unlikely
+to change in future).
```
yannikschaelte:
```suggestion
to change in the future).
```

petab/composite_problem.py
```diff
@@ -0,0 +1,75 @@
...
+        old_wd = os.getcwd()
+        if isinstance(yaml_config, str):
+            new_wd = os.path.dirname(yaml_config)
+            yaml_config = yaml.load_yaml(yaml_config)
+
+        yaml_config = yaml.load_yaml(yaml_config)
+
+        try:
+            if new_wd:
+                os.chdir(new_wd)
```
yannikschaelte: do you have to chdir? can potentially lead to problems e.g. when stuff fails, or when log files are written. can this be done by prefixing the locations instead?

petab/petab_schema.yaml
```diff
@@ -0,0 +1,46 @@
...
+$schema: "http://json-schema.org/draft-06/schema"
+description: PEtab parameter estimation problem config file schema
+
+properties:
+
+  petab_version:
+    type: string
+    description: Version of the PEtab format (e.g. 0.0.0a16)
+
+  parameter_file:
```
yannikschaelte: we discussed also allowing for multiple parameter files, and then doing a (n interactive) merge. one could have that as part of the standard, or have a helper function that creates a petab conform single parameter_file. opinions?

petab/petab_schema.yaml
```diff
@@ -0,0 +1,46 @@
...
+    description: |
+      One or more (SBML model, measurement, condition files) tuples.
+    items:
+      type: object
+      description: PEtab SBML, condition and measurement tables
+      properties:
+        sbml_file:
+          type: string
+          description: PEtab SBML file
+        measurement_files:
```
yannikschaelte: what is a use case for having multiple measurement files, but not multiple <rest>s?

petab/petab_schema.yaml
```diff
@@ -0,0 +1,46 @@
...
+          type: array
+          description: List of PEtab measurement files
+          items:
+            type: string
+            description: PEtab measurement file
```
yannikschaelte: analogous to measurement file, also specify the (optional) simulation file here? depends on whether we want to regard these as parts of the petab problem.

petab/petab_schema.yaml
yannikschaelte: in the prior case, there might be an alternative parameter_file field under problems.

petab/petab_schema.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  petab_version:
+    type: string
+    description: Version of the PEtab format (e.g. 0.0.0a16)
+
+  parameter_file:
+    type: string
+    description: |
+      PEtab parameter file name containing parameters of all models listed in
+      `problems`. Absolute or relative path.
```
yannikschaelte: I would directly allow general URLs here (including access mechanisms, e.g. http or file)

doc/documentation_data_format.md
dweindl: This was old, just moved two blocks. I think those extensions need some more explanation. For hierarchicalOptimization it is moderately unclear what happens. I think would remove that and put it rather into pyPESTO as "PEtab extensions". Maybe not in this PR though.

petab/composite_problem.py
dweindl: Probably rather exotic use case where this causes trouble (changing back at end of function) but will adapt.

petab/petab_schema.yaml
dweindl: I think we don't have that use case yet. I would keep it for a later version.

Merging parameter files: Strongly in favor of putting it in a separate function. Will leave implementation to first person having a use (and test) case for that. Will create an issue.

petab/petab_schema.yaml
dweindl: You may want to organize your measurements by origin of dataset and combine different subsets in different yaml files.
You are right though that the same argument may apply to condition tables too.
For SBML models, maybe as well. This will however require changes in a looooot of places, worth a separate PR.
I'd adapt the schema, so we don't need to change it too often, but raise NotYetImplemented.

petab/petab_schema.yaml
dweindl: "simulation file" the place to write results to? I think I'd keep that out of here.

petab/petab_schema.yaml
dweindl: Sounds useful. Will adapt the description. Implementation to be done at some later point. For the tables there shouldn't be any changes required except for file existance checks. For the SBML files, we'd have to implement it by ourselves.

petab/petab_schema.yaml
yannikschaelte: yes, I man the file that has the same format as measurement_file and contains the simulations. I think that exists now for various benchmark models, would be the question if we want to have it as optional for the standard now. point in favor would be that this can further automatize figure generation.

dweindl: Fair point. However, this may need some more thoughts. In case of multi-start optimization we may want to have multiple result files. The we need to allow for some kind of wildcards. This is again quite tool specific. Would tackle that later.

For the benchmark collection, where exactly one file is present, I'd handle that with a separate file with benchmark-specific info.

More important with respect to the visualization would be providing the visualization schema. Will leave that to people working on viz.

petab/petab_schema.yaml
yannikschaelte: background: this will in particular allow using distributed and online sources, as with the jws repo.

petab/petab_schema.yaml
dweindl: done

dweindl:
Thanks for comments. All done.

doc/documentation_data_format.md
```diff
@@ -366,9 +366,42 @@ Additional columns may be added.
...
+- `initializationPriorParameters` (optional)
+
+  Prior parameters used for sampling of initial point for optimization. Uses
+  the entries from `priorParameters` as default, but will overwrite those, if
+  something else is specified here. For more detailed documentation, see
+  `priorParameters`.
+
+- `objectivePriorType` (optional)
+
+  Prior types used for the objective function during optimization. Uses the
```
yannikschaelte: Note: This prior is presumably not only used in optimization, but also in sampling.

doc/documentation_data_format.md
```diff
@@ -469,44 +502,31 @@ order:
...
-
-- `objectivePriorParameters` (optional)
-
-  Prior parameters used for the objective function during optimization. Uses
-  the entries from `priorParameters` as default, but will overwrite those, if
-  something else is specified here. For more detailed documentation, see
-  `priorParameters`.   
+Parameter estimation problems can comprise multiple models. For now, PEtab
+allows to specify multiple SBML models with corresponding condition and
+measurement tables, and one joint parameter table. This means that parameter
```
yannikschaelte:
```suggestion
measurement tables, and one joint parameter table. This means that the parameter
```

petab/composite_problem.py
```diff
@@ -0,0 +1,77 @@
...
+            problems: List[problem.Problem] = None):
+        """Constructor
+
+        Arguments:
+            parameter_df:
+                see CompositeProblem.parameter_df
+            problems:
+                see CompositeProblem.problems
+        """
+        self.problems: List[problem.Problem] = problems
```
yannikschaelte: didn't know one can add type info here in python ...

petab/composite_problem.py
```diff
@@ -0,0 +1,77 @@
...
+        self.parameter_df: pd.DataFrame = parameter_df
+
+    @staticmethod
+    def from_yaml(yaml_config: Union[Dict, str]
+                  ) -> 'CompositeProblem':
```
yannikschaelte: this is ugly format. + can one use the class name instead of str for ret type?

petab/composite_problem.py
yannikschaelte: doesn't seem adapted yet, but should be fine

petab/composite_problem.py
```diff
@@ -0,0 +1,77 @@
...
+            parameter_df = parameters.get_parameter_df(
+                yaml_config['parameter_file'])
+            problems = []
+            for problem_config in yaml_config['problems']:
+                yaml.assert_single_condition_and_sbml_file(problem_config)
+
+                # don't set parameter file if we have multiple models
+                cur_problem = problem.Problem.from_files(
+                    sbml_file=problem_config['sbml_files'][0],
+                    measurement_file=problem_config['measurement_files'],
```
yannikschaelte: so the current plan is to in the code support only a single sbmle and condition file, but several measrurement_files?
one "un-nicety" is that the variable measurement_file is singular.

petab/core.py
```diff
@@ -90,7 +90,7 @@ def get_observable_id(parameter_id: str) -> str:
 
 
 def flatten_timepoint_specific_output_overrides(
-        petab_problem: problem.Problem) -> None:
+        petab_problem: 'problem.Problem') -> None:
```
yannikschaelte: this is to avoid circular imports?

petab/lint.py
```diff
@@ -371,16 +368,17 @@ def assert_parameter_estimate_is_boolean(parameter_df: pd.DataFrame) -> None:
...
     Are there time-point or replicate specific parameter assignments in the
     measurement table.
 
     Arguments:
         measurement_df: PEtab measurement table
 
-    Raises:
-        AssertionError: in case of problems
+    Returns:
+        True, if there time-point or replicate specific parameter assignments
```
yannikschaelte:
```suggestion
        True if there are time-point or replicate specific parameter
```

petab/lint.py
```diff
@@ -371,16 +368,17 @@ def assert_parameter_estimate_is_boolean(parameter_df: pd.DataFrame) -> None:
...
+        in the measurement table, False otherwise.
```
yannikschaelte:
```suggestion
        assignments in the measurement table, False otherwise.
```

petab/composite_problem.py
yannikschaelte: but no problem here since it is flattened right away.

petab/petab_schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+      required:
+        - sbml_files
+        - measurement_files
+        - condition_files
+
+required:
+  - petab_version
+  - parameter_file
+  - problems
```
yannikschaelte: looks good now, we will need to implement it on one model (Becker e.g.) to see whether it holds.

petab/petablint.py
```diff
@@ -82,9 +86,15 @@ def parse_cli_args():
                 folder=args.directory,
             )
 
-    if not args.model_name and not args.sbml_file_name \
-        and not args.condition_file_name and not args.measurement_file_name \
-            and not args.parameter_file_name:
+    if (args.yaml_file_name
+            and (args.sbml_file_name or args.condition_file_name
```
yannikschaelte: and any([...]) might be nicer here (no opinion though)

petab/petablint.py
```diff
@@ -82,9 +86,15 @@ def parse_cli_args():
...
+                 or args.measurement_file_name or args.parameter_file_name)):
+        parser.error('When providing a yaml file, no other files may '
+                     'be specified.')
+
+    if (not args.model_name and not args.sbml_file_name
```
yannikschaelte: if not any([...]) ...

petab/petablint.py
```diff
@@ -103,26 +113,38 @@ def main():
...
-        )
-    except FileNotFoundError as e:
-        logger.error(e)
-        sys.exit(1)
+    if args.yaml_file_name:
+        if petab.is_composite_problem(args.yaml_file_name):
+            from petab.yaml import validate
+            validate(args.yaml_file_name)
+
+            # TODO: further checking
```
yannikschaelte: TODO Issue?

petab/problem.py
```diff
@@ -109,6 +115,41 @@ def from_files(sbml_file: str = None,
...
+        if isinstance(yaml_config, str):
+            new_wd = os.path.dirname(yaml_config)
+            yaml_config = yaml.load_yaml(yaml_config)
+
+        if yaml.is_composite_problem(yaml_config):
+            raise ValueError('petab.Problem.from_yaml() can only be used for '
+                             'yaml files comprising a single model. '
+                             'Consider using '
+                             'petab.CompositeProblem.from_yaml() instead.')
+        try:
```
yannikschaelte: for future reference, this repeating code pattern might be nice to address with a context manager.

petab/sampling.py
```diff
@@ -5,7 +5,7 @@
 
 from typing import Tuple
 
-from .parameters import get_priors_from_df
+from. import parameters
```
yannikschaelte:
```suggestion
from . import parameters
```

petab/yaml.py
```diff
@@ -0,0 +1,150 @@
...
+
+def assert_single_condition_and_sbml_file(problem_config: Dict) -> None:
+    """Check that there is only a single condition file and a single SBML
+    file specified.
+
+    Arguments:
+        problem_config:
+            Dictionary as defined in the YAML schema inside the `problems`
+            list.
+    Raises:
```
yannikschaelte: nothing?

petab/composite_problem.py
dweindl: Not inside the class, because not yet known at that point.

petab/core.py
dweindl: yes, pytest had troubles for reasons i didn't fully understand

petab/petab_schema.yaml
dweindl: Files are generated for all models: https://github.com/LeonardSchmiester/Benchmark-Models/pull/84

petab/petablint.py
dweindl: Will consider it next time ;)

petab/petablint.py
dweindl: okay, okay...

petab/problem.py
dweindl: Right. But now removed anyways...

petab/composite_problem.py
dweindl: I also noticed that, but didn't find it severe enough to justify plenty of downstream changes. Not opposed to changing it though. Maybe along with all the others after #188 #189.

PR #198, opened 2019-12-17
Add Problem properties to retrieve scaled parameters; add function to…

dweindl:
… scale parameters

### DIFF
--- a/petab/parameters.py
+++ b/petab/parameters.py
@@ -232,3 +232,28 @@ def parameter_id_is_valid(parameter_id: str) -> bool:
     """
 
     return parameter_id != ''
+
+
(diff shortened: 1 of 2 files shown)


petab/parameters.py
```diff
@@ -232,3 +232,28 @@ def parameter_id_is_valid(parameter_id: str) -> bool:
...
+
+
+def scale(parameter: numbers.Number, scale_str: 'str') -> numbers.Number:
+    """Scale parameter according to scale_str
+
+    Arguments:
+        parameter:
+            Parameter to be scaled
+        scale_str:
+            One of '', 'lin', 'log', 'log10'
```
yannikschaelte: its repetitive, but please say again that '' becomes 'lin'

PR #203, opened 2019-12-18
Fix missing brackets resulting in `petablint` error

dweindl:


petab/petablint.py
```diff
@@ -93,9 +93,9 @@ def parse_cli_args():
                      'be specified.')
 
     if (not args.model_name
-            and not any(args.sbml_file_name, args.condition_file_name,
+            and not any([args.sbml_file_name, args.condition_file_name,
```
yannikschaelte: 🤣 yes, any does not make use of *args

PR #211, opened 2020-01-14
Enable passing dataframes to visualization functions as an alternative to file paths

plakrisenko:
improvements related to #162 
possibilities are added to visualize petab problem directly and to pass dataframes to the visualization function as an alternative to file paths


petab/visualize/plot_data_and_simulation.py
```diff
@@ -28,14 +34,15 @@ def plot_data_and_simulation(data_file_path: str,
 
     Parameters
     ----------
-    data_file_path: str
-        Path to the data file.
-    condition_file_path: str
-        Path to the condition file.
+    exp_data: str
```
dweindl:
```suggestion
    exp_data:
```

As you added typehints (:+1:), it would be good to remove the datatypes from the docstrings. Otherwise we need to update two places. And they are already out of sync ;).

plakrisenko: yep, you're right

Issue #205, opened 2019-12-18
Parallelize petab.get_optimization_to_simulation_parameter_mapping


dweindl:
Takes too long for large models and is already in good shape to generate mapping for different simulation conditions in parallel.

PR #215, opened 2020-01-15
Parallelize petab.get_optimization_to_simulation_parameter_mapping (C…

dweindl:
…loses #205)


petab/parameter_mapping.py
```diff
@@ -63,36 +69,68 @@ def get_optimization_to_simulation_parameter_mapping(
...
-                condition_id=condition.preequilibrationConditionId,
-                is_preeq=True,
-                cur_measurement_df=cur_measurement_df,
-                condition_df=condition_df,
-                parameter_df=parameter_df, sbml_model=sbml_model,
-                warn_unmapped=warn_unmapped
-            )
+    num_processes = 1
+    if ENV_NUM_THREADS in os.environ:
+        num_processes = int(os.environ[ENV_NUM_THREADS])
```
yannikschaelte: ``num_processes = int(os.environ.get(ENV_NUM_THREADS, 1)``

petab/parameter_mapping.py
```diff
@@ -63,36 +69,68 @@ def get_optimization_to_simulation_parameter_mapping(
...
+
+    from multiprocessing import Pool
```
yannikschaelte: don't like in-code imports, or is that a useful thing?

petab/parameter_mapping.py
```diff
@@ -63,36 +69,68 @@ def get_optimization_to_simulation_parameter_mapping(
...
+    pool = Pool(processes=num_processes)
```
yannikschaelte: so now you perform the map in all cases, using multiprocessing. but that pickles the objects, right? so for 1 process, the default, you get some overhead?

yannikschaelte: would multithreading be an option?

petab/parameter_mapping.py
```diff
@@ -134,19 +179,24 @@ def get_parameter_mapping_for_condition(
...
     _perform_mapping_checks(cur_measurement_df)
 
-    par_sim_ids = sbml.get_model_parameters(sbml_model)
+    if simulation_parameter_ids is not None and sbml_model is None:
+        pass
+    elif simulation_parameter_ids is None and sbml_model is not None:
+        simulation_parameter_ids = sbml.get_model_parameters(sbml_model)
+    else:
+        raise ValueError("Must provide exactly one of `sbml_model` and "
+                         "`simulation_parameter_ids`.")
```
yannikschaelte: can be shortened: if pid is None and model is not None: ...; elif (pid is None) == (model is None): Error

tests/test_parameter_mapping.py
```diff
@@ -138,6 +138,13 @@ def test_partial_override(condition_df_2_conditions,
             sbml_model=sbml_model
         )
 
+        # Comparison with NaN containing expected results fails after pickling!
```
yannikschaelte: also add a test with more than one core?

petab/__init__.py
```diff
@@ -1,4 +1,13 @@
-"""PEtab exports"""
+"""PEtab exports
+
+Attributes:
+    ENV_NUM_THREADS:
+        Name of environment variable to set number of threads or processes
```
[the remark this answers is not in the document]
dweindl: Preferring less 'the's

petab/parameter_mapping.py
dweindl: Generally agreed. But maybe there is no need to import we only use it in this one specific location, which is usually not called more than once.

petab/parameter_mapping.py
dweindl: Overhead, yes. Could add special treatment. Although for that case, not sure if the overhead is worse than adding more complexity. Multi-threading: Possibly.

petab/parameter_mapping.py
dweindl: You are right, but found the current way easier to read.

tests/test_parameter_mapping.py
dweindl: Can do, although it's a bit hard to check anything besides not-crashing. Therefore, I didn't see too much value. If sequential gets a separate implementation as suggested above, then it would make more sense. Although I'd rather avoid that.

petab/parameter_mapping.py
yannikschaelte: ok, I guess pickling only becomes a problem with large problems, but then you will wanna parallelize anyway, so should be fine

tests/test_parameter_mapping.py
yannikschaelte: yes, I was just saying check that single-core and two-core gives the same result, nothing more (could e.g. be that pool has a special treatment for 1 process)

dweindl: Ah, right, that would make sense.

petab/parameter_mapping.py
dweindl: threads and special handling of sequential -> done

PR #222, opened 2020-01-17
Update prior definitions; add checks

### LINKED ISSUE #218: Parameter file: Simplify prior specification
Continuation of #17, see also #131, #206.

Currently, there are {priorX}, {initializationPriorX}, {objectivePriorX}, with X = {Type|Parameters}. The sole prior is supposed to simplify the setting when intitialization and objective prior coincide. However, this leads to unclear overwriting. Thus, it seems preferable to only keep initialization and objective, i.e. enforce defining both. While this creates some overhead when both coincide, it makes things clearer and less error-prone.

yannikschaelte:
Fixes #218


doc/documentation_data_format.md
```diff
@@ -326,13 +326,11 @@ Additional columns may be added.
...
-  Type of prior, which is used for sampling of initial points for 
-  a possible optimization and for the objective function. Priors which are 
-  only used for sampling of initial starting points or only for optimization
-  should be specified in the additional columns `initializationPriorType` or
-  `objectivePriorType`, respectively.
+  Prior types used for sampling of initial points for optimization. Sampled
+  points are clipped to lie inside the parameter boundaries specified by
+  `lowerBound` and `upperBound`. Defaults to `parameterScaleUniform`.
 
   Possible prior types are (see also Extensions):
```
dweindl:
```suggestion
  Possible prior types are:
```

"extension" is gone now, right?

doc/documentation_data_format.md
```diff
@@ -365,43 +361,25 @@ Additional columns may be added.
...
+  Prior parameters used for the objective function during optimization.
+  For more detailed documentation, see `initializationPriorParameters`.   
 
-  Prior parameters used for sampling of initial points for optimization. Uses
-  the entries from `priorParameters` as default, but will overwrite those, if
-  something else is specified here. For more detailed documentation, see
-  `priorParameters`.
 
-- `objectivePriorType` (optional)
+### Additional optional columns
```
dweindl: Not related to sampling, but I think we wanted to kick that out here, as that will be quite tool-specific.

doc/documentation_data_format.md
yannikschaelte: yes

yannikschaelte:
note: a check on the priorType column was added, but priorParameters is not checked yet.

yannikschaelte:
currently, we write e.g. `uniform  3;3` and `parameterScaleUniform` assuming log10 parameters and bounds 1e-8, 1e8. It needs to be specified on what scale those priorParameters are. Would it make sense to have them on lin scale for uniform, and log10 scale for parameterScaleUniform?

yannikschaelte:
> currently, we write e.g. `uniform -3;3` and `parameterScaleUniform -3;3` assuming log10 parameters and bounds 1e-8, 1e8. It needs to be specified on what scale those priorParameters are. Would it make sense to have them on lin scale for uniform, and log10 scale for parameterScaleUniform?

done

PR #253, opened 2020-02-01
Add check for valid identifiers (Closes #179)

### LINKED ISSUE #179: Add check for valid parameter IDs
This should pretty much correspond to what is allowed for SBML identifiers

In `petab.parameter.parameter_id_is_valid`

dweindl:


plakrisenko:
What about 
`re.match(r'^[a-zA-Z]|(_+[a-zA-Z\d])\w*$', x)`
to avoid only underscores?

dweindl:
> What about
> `re.match(r'^[a-zA-Z]|(_+[a-zA-Z\d])\w*$', x)`
> to avoid only underscores?

Agreed that only underscores would make pretty bad IDs. On the other hand, _1, _2, ... are probably not much better. Only underscores are allowed e.g. in SBML. Would only exclude things that cause technical problems later on, and have some trust in the sanity of the user.

petab/lint.py
```diff
@@ -697,7 +697,7 @@ def is_valid_identifier(x: str) -> bool:
 
     Check whether `x` is a valid identifier for conditions, parameters,
     observables... . Identifiers may contain upper and lower case letters,
-    digits and underscores, but must not start with a letter.
+    digits and underscores, but must not start with a digit.
```
dweindl: :+1:

PR #256, opened 2020-02-06
allow to extract only estimate parameters

yannikschaelte:


yannikschaelte:
not sure how to call them best

PR #268, opened 2020-02-06
Add properties for fixed/free and scaled values

yannikschaelte:


petab/problem.py
```diff
@@ -346,13 +346,19 @@ def x_ids(self) -> List[str]:
         """Parameter table parameter IDs"""
         return list(self.parameter_df.reset_index()[PARAMETER_ID])
 
+    @property
+    def x_free_ids(self) -> List[str]:
+        """Parameter table parameter IDs, for the parameters to estimate."""
+        estimated = self.parameter_df[self.parameter_df[ESTIMATE] != 0]
+        return list(estimated.reset_index()[PARAMETER_ID])
```
dweindl:
```suggestion
        return list(estimated.index.values)
```

PR #289, opened 2020-02-13
Fix format documentation regarding parameter scale

dweindl:


dweindl:
Yeah, sorry for breaking things there. Apart from the breaking change, I think it makes life easier in the future though. If one only wants to simulate the model, one can simply ignore the parameter scaling, and it put an end to the confusion with the observable transformation which was handled incoherently.

PR #335, opened 2020-02-23
Fix links in docs

yannikschaelte:
replace relative by absolute paths when creating module.


yannikschaelte:
> Great!
> Assuming it works.
We will find out :) locally yes
> Also of interest for AMICI and pyPESTO.
Yes

PR #347, opened 2020-02-27
Allow parametric noise overrides in noise formula for residuals

yannikschaelte:


yannikschaelte:
@dweindl does this capture all replacement variants we allow?

petab/calculate.py
```diff
@@ -95,16 +105,81 @@ def calculate_residuals_for_table(
...
+
+    Returns:
+        noise_formulas: Dictionary of {observable_id}: {noise_formula}.
+    """
+    noise_formulas = {}
+    # iterate over observables
+    for row in observable_df.itertuples():
+        observable_id = row.Index
+        if NOISE_FORMULA not in observable_df.columns:
+            noise_formula = []
```
dweindl: Is empty list preferable over None?

petab/calculate.py
```diff
@@ -12,6 +14,7 @@ def calculate_residuals(
         measurement_dfs: Union[List[pd.DataFrame], pd.DataFrame],
         simulation_dfs: Union[List[pd.DataFrame], pd.DataFrame],
         observable_dfs: Union[List[pd.DataFrame], pd.DataFrame],
+        parameter_df: Union[List[pd.DataFrame], pd.DataFrame],
```
dweindl:
```suggestion
        parameter_dfs: Union[List[pd.DataFrame], pd.DataFrame],
```
Consistency with other arguments?

yannikschaelte: ah right. that would depend also on if we now allow an array of parameter dfs in yaml.

petab/calculate.py
yannikschaelte: no

PR #352, opened 2020-02-28
Fix validator: nominalValue is optional unless estimated!=1 anywhere

dweindl:


petab/lint.py
```diff
@@ -166,6 +166,23 @@ def check_parameter_df(
...
+    non_estimated_par_ids = list(df.index[(df[ESTIMATE] != 1)
+                                          | (df[ESTIMATE] != '1')])
+    if non_estimated_par_ids:
+        if NOMINAL_VALUE not in df:
+            raise AssertionError("Parameter table contains parameters "
+                                 f"{non_estimated_par_ids} that are not "
+                                 "specified to be estimated, "
+                                 f"but column {NOMINAL_VALUE} is missing.")
+        try:
+            df.loc[non_estimated_par_ids, NOMINAL_VALUE].apply(float)
```
yannikschaelte:
```suggestion
            df.loc[non_estimated_par_ids, NOMINAL_VALUE].apply(int)
```

yannikschaelte: ist ``in [0,1]`` explicitly checked anywhere?

dweindl: yes `assert_parameter_estimate_is_boolean`

dweindl: But this one is about nominalValue, not about estimate. So float seems more appropriate.

PR #365, opened 2020-03-04
Update parameter mapping to deal with species and compartments in condition table

dweindl:


petab/parameter_mapping.py
```diff
@@ -336,8 +336,15 @@ def _apply_condition_parameters(par_mapping: ParMappingDict,
...
 
+        # Species and compartments are handled elsewhere
+        if sbml_model.getSpecies(overridee_id) is not None:
+            continue
+        if sbml_model.getCompartment(overridee_id) is not None:
+            continue
+
         par_mapping[overridee_id] = core.to_float_if_float(
             condition_df.loc[condition_id, overridee_id])
+        scale_mapping[overridee_id] = LIN
```
yannikschaelte: yes, should work. we assume that all values are in lin space, and for condition specific numeric overrides, we will never need scales, as these are not estimated.

dweindl: Yes, that was so far incoherent. Scale mappings had potentially fewer entries than parameter value mapping.

PR #366, opened 2020-03-05
fix observable dependant calculations of llh, chi2

yannikschaelte:
state overrides are currently not possible yet, because for that we would need all system states somehow available.


petab/calculate.py
```diff
@@ -165,6 +166,10 @@ def evaluate_noise_formula(
     for i_obs_par, obs_par in enumerate(observable_parameter_overrides):
         overrides[f"noiseParameter{i_obs_par+1}_{observable_id}"] = obs_par
 
+    # fill in observables
+    for row in simulation_df.itertuples():
+        overrides[row.observableId] = row.simulation
```
dweindl: petab.C.OBSERVABLE_ID? :)

yannikschaelte:
@dweindl should be fixed now

PR #367, opened 2020-03-05
Fix merge_preeq_and_sim_pars_condition

dweindl:
... which was for some reason assuming we have lists instead of dicts and was still mysteriously working in most cases.

### DIFF
--- a/petab/parameter_mapping.py
+++ b/petab/parameter_mapping.py
@@ -492,17 +492,29 @@ def merge_preeq_and_sim_pars_condition(
         # nothing to do
         return
 
-    for idx, (par_preeq, par_sim, scale_preeq, scale_sim) \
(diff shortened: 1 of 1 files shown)


petab/parameter_mapping.py
```diff
@@ -492,39 +492,55 @@ def merge_preeq_and_sim_pars_condition(
...
+    all_par_ids = set(condition_map_sim.keys()) \
+        | set(condition_map_preeq.keys())
+
+    for par_id in all_par_ids:
+        if par_id not in condition_map_preeq:
+            # nothing to do
+            continue
+
+        if par_id not in condition_map_sim:
+            # unmapped for simulation -> just add preeq
```
yannikschaelte:
```suggestion
            # unmapped for simulation -> just use preeq values
```

PR #369, opened 2020-03-06
Allow and require output parameters defined in observable table to be…

### LINKED ISSUE #361: Validator: False positive: "Extraneous parameter(s) in parameter table"
... occurring for parameter introduced in observables table.

dweindl:
… defined in parameter table

* Rearrange functions to correct modules
* Fixes #361
* Breaks petab.migrations.sbml_observables_to_table, therefore removing here


petab/core.py
```diff
@@ -366,3 +366,17 @@ def _add_file_metadata(location: str, description: str = ""):
...
+    occurrence of each element
+
+    seq: Sequence to prune
+
+    Returns:
+        List of unique elements in ``seq``
+    """
+    seen = set()
+    seen_add = seen.add
+    return [x for x in seq if not (x in seen or seen_add(x))]
```
yannikschaelte: funny but should work

petab/observables.py
```diff
@@ -79,3 +81,65 @@ def get_output_parameters(observable_df: pd.DataFrame,
...
+            is for observable or for noise model
+
+    Returns:
+        List of placeholder parameter IDs in the order expected in the
+        observableParameter column of the measurement table.
+    """
+    if not formula_string:
+        return []
+
+    if not isinstance(formula_string, str):
```
yannikschaelte: does this happen? could be dangerous if only "string-like" objects are passed. this is intended to test for numerics, right? maybe then test explicitly.

petab/observables.py
```diff
@@ -79,3 +81,65 @@ def get_output_parameters(observable_df: pd.DataFrame,
...
+
+    return placeholders
+
+
+def get_placeholders(observable_df: pd.DataFrame) -> List[str]:
+    """Get all placeholder parameters from observable table observableFormulas
+    and noiseFormulas
+
+    Arguments:
+        observable_df: PEtab observable table
```
yannikschaelte: returns?

petab/core.py
dweindl: it does :sweat_smile: . was there before, just moved and documented.

petab/observables.py
dweindl: Typehint asks for `str`. Not expecting anything funny then :).

PR #381, opened 2020-03-13
Deprecate petab.sbml.globalize_parameters

dweindl:
Function has nothing to do with PEtab's objectives, and is not usable
at the moment. Better functionality is available through
libsbml.SBMLLocalParameterConverter. Therefore, this should be removed.

### DIFF
--- a/petab/sbml.py
+++ b/petab/sbml.py
@@ -131,6 +131,10 @@ def globalize_parameters(sbml_model: libsbml.Model,
             Prepend reaction id of local parameter when
             creating global parameters
     """
+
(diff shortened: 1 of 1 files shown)


dweindl:
> please add note (e.g. in pr description) why this is deprecated now -- makes it easier to follow later.

"It should have never gotten in here"? :)

yannikschaelte:
> > please add note (e.g. in pr description) why this is deprecated now -- makes it easier to follow later.
> 
> "It should have never gotten in here"? :)

exactly :) now it's documented ;)

PR #408, opened 2020-03-24
Create datasetId also for simulation dataframe (resolves #388)

### LINKED ISSUE #388: visualization bug - without visualizationSpecification
When plotting __without visualizationSpecification__ file if `datasetId` is not present as column name in simulatedData files executing the following returns an error
``` python
plot_data_and_simulation(
    exp_data=data_file_path, exp_conditions=condition_file_path,sim_data=simulation_file_path,
    sim_cond_id_list = [['model1_data1']])
```
``` python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-37-362d13aee1cf> in <module>
     20 plot_data_and_simulation(
     21     exp_data=data_file_path, exp_conditions=condition_file_path,sim_data=simulation_file_path,
---> 22     sim_cond_id_list = [['model1_data1']])
     23 
     24 plt.show()

/usr/local/lib/python3.7/dist-packages/petab-0.1.4-py3.7.egg/petab/visualize/plot_data_and_simulation.py in plot_data_and_simulation(exp_data, exp_conditions, vis_spec, sim_data, dataset_id_list, sim_cond_id_list, sim_cond_num_list, observable_id_list, observable_num_list, plotted_noise, subplot_file_path)
    137 
    138     if sim_data is not None and DATASET_ID not in sim_data:
--> 139         raise ValueError(f'Visualization requires field {DATASET_ID} to be '
    140                          f'present in simulation table.')
    141 

ValueError: Visualization requires field datasetId to be present in simulation table.
```
However this only occurs when using simulated data. If only plotting measurementData (and no `datasetId` present) it works.

After including the column name `datasetId` all works.

So far, I thought it is not mandatory to have this field in the simulatedData table.
- or is it?

MerktSimon:
Function check_ex_exp_columns is changed to also check simulation columns and create datasetId column in case it is missing

### DIFF
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -201,12 +201,11 @@ def create_dataset_id_list(
         # create nicer legend entries from condition names instead of IDs
         if dataset_id not in legend_dict.keys():
             tmp = exp_conditions.loc[exp_conditions.index == cond_id]
-            try:
-                legend_dict[dataset_id] = tmp.conditionName[0] + ' - ' + \
-                    tmp_obs[ind]
-            except (AttributeError, np.UFuncTypeError):
-                legend_dict[dataset_id] = tmp.index[0] + ' - ' + \
-                    tmp_obs[ind]
+            if CONDITION_NAME not in tmp.columns or tmp[
+                    CONDITION_NAME].isna().any():
+                tmp.loc[:, CONDITION_NAME] = tmp.index.tolist()
+            legend_dict[dataset_id] = tmp[CONDITION_NAME][0] + ' - ' + \
+                tmp_obs[ind]
 
     # add these column to the measurement table (possibly overwrite)
     if DATASET_ID in exp_data.columns:
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -433,7 +433,8 @@ def check_ex_exp_columns(
         raise NotImplementedError(
             "Column \'simulationConditionId\' is missing in measurement "
             "file. ")
-    if MEASUREMENT not in exp_data.columns:
+    if MEASUREMENT not in exp_data.columns and SIMULATION not in \
+            exp_data.columns:
         raise NotImplementedError(
             "Column \'measurement\' is missing in measurement "
             "file. ")
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -412,6 +412,8 @@ def check_ex_visu_columns(vis_spec: pd.DataFrame,
         vis_spec[X_LABEL] = 'time'
     if X_OFFSET not in vis_spec.columns:
         vis_spec[X_OFFSET] = 0
+    if Y_OFFSET not in vis_spec.columns:
+        vis_spec[Y_OFFSET] = 0
     if Y_SCALE not in vis_spec.columns:
         vis_spec[Y_SCALE] = LIN
     if X_SCALE not in vis_spec.columns:
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -204,7 +204,7 @@ def create_dataset_id_list(
             try:
                 legend_dict[dataset_id] = tmp.conditionName[0] + ' - ' + \
                     tmp_obs[ind]
-            except AttributeError:
+            except (AttributeError, np.UFuncTypeError):
                 legend_dict[dataset_id] = tmp.index[0] + ' - ' + \
                     tmp_obs[ind]
 
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -29,20 +29,25 @@ NumList = List[int]
 def import_from_files(
         data_file_path: str,
         condition_file_path: str,
-        visualization_file_path: str,
         simulation_file_path: str,
         dataset_id_list: List[IdsList],
         sim_cond_id_list: List[IdsList],
         sim_cond_num_list: List[NumList],
         observable_id_list: List[IdsList],
         observable_num_list: List[NumList],
-        plotted_noise: str) -> Tuple[pd.DataFrame, pd.DataFrame,
-                                     pd.DataFrame, pd.DataFrame]:
+        plotted_noise: str,
+        visualization_file_path: str = None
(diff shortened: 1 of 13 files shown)


elbaraim:
@MerktSimon but is not `datasetId` an optional field? why would you create it?

petab/visualize/helper_functions.py
```diff
@@ -468,7 +468,8 @@ def check_ex_exp_columns(
         raise NotImplementedError(
             "Column \'simulationConditionId\' is missing in measurement "
             "file. ")
-    if MEASUREMENT not in exp_data.columns:
+    if MEASUREMENT not in exp_data.columns and SIMULATION not in \
```
erikadudki: this only allows the case if measurement & simulation is not available, is this intended? Can this happen? Isnt it either measurement is not available or simulation?

petab/visualize/helper_functions.py
```diff
@@ -468,7 +468,8 @@ def check_ex_exp_columns(
...
+            exp_data.columns:
         raise NotImplementedError(
             "Column \'measurement\' is missing in measurement "
             "file. ")
```
erikadudki: maybe adding in the error, that also simulation is missing.

petab/visualize/helper_functions.py
erikadudki: and for the measurement data, 'simulation' is not needed?

petab/visualize/plot_data_and_simulation.py
```diff
@@ -100,7 +100,7 @@ def plot_data_and_simulation(
         # import from file
         exp_data = measurements.get_measurement_df(exp_data)
         # check columns, and add non-mandatory default columns
-        exp_data, dataset_id_list, legend_dict = \
+        exp_data, _, legend_dict = \
```
erikadudki: could you explain this change? is dataset_id_list not necessary anymore?

petab/visualize/helper_functions.py
MerktSimon: a dataframe has to have eiter `simulation` or `meaurement. This is a cheap way to make the function validate measument dfs and simulation dfs at the same time

MerktSimon:
@elbaraim it is creating the `datasetId` column only for the tables loaded as dataframe in order to continue with visualization. the local files stay untouched

petab/visualize/plot_data_and_simulation.py
MerktSimon: I think it never was in reasonable cases:
* if you hand over to plot_data_and_simulation a dataset_id_list of ids you want to plot, then this list is left untouched by the check_ex_exp_col function
* if you hand over something else like cond_id_list to plot_data_and_simulation then you do not have a visu_df. In line 103 then exp_data gets a new dataset_id column which is also needed. Later in the code a virtual visu_df is created with a dataset_id column made in the same way
* if you provide visu_df than it hast mandatory the dataset_id column and so should mes_df and sim_df have

erikadudki: thanks for the explanation! okay, the only thing I'm not sure at the moment, I think we decided that dataset_id column is not mandatory anymore, also if you provide visu_df... but I dont know if this was already addressed.... but maybe this is then another issue

MerktSimon: ok, you are right. It is not mandatory anymore. Thanks, I missed that. I will change the code.

petab/visualize/helper_functions.py
MerktSimon: I will give the check_ex_exp_col function an option for simulation to make it more clean

elbaraim:
@MerktSimon would be possible to silent this?
```
{'plot1': <matplotlib.axes._subplots.AxesSubplot at 0x7ff792722110>,
 'plot2': <matplotlib.axes._subplots.AxesSubplot at 0x7ff7926a0b50>,
 'plot3': <matplotlib.axes._subplots.AxesSubplot at 0x7ff79264a390>}
```
I think by adding `plt.show()` would work.

MerktSimon:
@elbaraim this is actually a feature not a bug.
`plot_data_and_simulation()` returns a dictionary of axes objects so that users can chose what to do with them.
You can silent this by e.g. `ax = plot_plot_data_and_simulation(...`

elbaraim:
absolutely! :sweat_smile:

petab/visualize/plot_data_and_simulation.py
```diff
@@ -125,18 +139,8 @@ def plot_data_and_simulation(
...
-
-    if DATASET_ID not in exp_data:
-        raise ValueError(f'Visualization requires field {DATASET_ID} to be  '
-                         f'present in measurement table.')
-
-    if sim_data is not None and DATASET_ID not in sim_data:
-        raise ValueError(f'Visualization requires field {DATASET_ID} to be '
-                         f'present in simulation table.')
+        if sim_data is not None:
+            sim_data[DATASET_ID] = exp_data[DATASET_ID]
```
FFroehlich: I feel like this line is just asking for trouble. It might work for that one specific use case but is generally not a good solution.
- you don't want to overwrite user provided value
- I don't think we can generally assume that sim_data and exp_data are equally ordered.

Also I generally recommend that if you fix something like this, add one or two test cases to make sure it is not inadvertently broken in the future.

MerktSimon: * `if sim_data is not None and DATASET_ID not in sim_data:` Hence this only applies if there is no user provided datasetId. In these cases exp_data gets a default datasetId (something like 'obsverbale_condition').
* I thought in cases without datasetId it would be reasonable to assume same order for sim_data and exp_data. But of course the procedure for creating exp_data's datasetId could be easily applied to sim_data, too.

MerktSimon: test wise this could be covered by enhancing each of the plot_by_observable/condition/datasetId tests with a second function call also plotting with simulated data and not only measurement

FFroehlich: > * `if sim_data is not None and DATASET_ID not in sim_data:` Hence this only applies if there is no user provided datasetId. In these cases exp_data gets a default datasetId (something like 'obsverbale_condition').

I can only see `if sim_data is not None` there. As a proposal for a fix this sounds right.

MerktSimon: ah yeah you are right. had to many intermediate versions.

PR #409, opened 2020-03-25
fix arg type

yannikschaelte:


yannikschaelte:
in amici we fill condition from df.iterrows() -> Iterable[int, Series], so I guess yes

Issue #415, opened 2020-03-26
tests minimal visualization table


MerktSimon:
The tests for minimal visualization tables are currently switched off since the data format was incoherent with respect to `datasetId` column (cf late discussion of #341).
Because `datasetId` is now optinal in the visu table the tests should now work

Issue #272, opened 2020-02-07
Add commented example for creation of visualization files


dweindl:

yannikschaelte:
closed via #423

Issue #190, opened 2019-12-16
Implement validation of PEtab visualization files


dweindl:
- [ ] Implement consistency checks
  - [ ] plotId
  - [ ] [plotName]
  - [ ] plotTypeSimulation
  - [ ] plotTypeData
  - [ ] datasetId
  - [ ] [xValues]
  - [ ] [xOffset]
  - [ ] [xLabel]
  - [ ] [xScale]
  - [ ]  [yValues]
  - [ ] [yOffset]
  - [ ] [yLabel]
  - [ ] [yScale]
  - [ ] [legendEntry]
- [ ] Add as command line option to `petablint`

MerktSimon:
also cross checks like `barplot` in `plotType` if `condition` in `xValues` should be implemented

PR #418, opened 2020-03-26
more tests for visualization routine

MerktSimon:
activate tests for minimal data file (cf issue #415)
tests without visu file now also with simu data

One might still argue that the minimal visualization file is not minimal since it still has the datasetId column. Imo it is only optional if one provide a dataset, condition or observable list to the function plot_data_and_simulation()
Any opinions on this?

### DIFF
--- a/petab/visualize/helper_functions.py
+++ b/petab/visualize/helper_functions.py
@@ -59,14 +59,14 @@ def import_from_files(
         vis_spec = petab.get_visualization_df(visualization_file_path)
     else:
         # create them based on simulation conditions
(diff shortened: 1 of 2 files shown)


JanHasenauer:
Very good idea! The test would be very useful.

MerktSimon:
Right now the routine seems still to struggle withouth the datasetId column.
But @plakrisenko  is currently working on #272 and #190 including support of this. (https://github.com/PEtab-dev/PEtab/tree/pl_visualization)
I will include the according tests there.

Issue #428, opened 2020-04-21
Visualization if XValue Column not provided and multiple conditions should be plotted


erikadudki:
If in the visualization table, the column XValues is not provided, it is inferred:
- if in measurement file measurement time is not constant, then XValues = 'time' ->plot  time course
- if its constant, then dose response plots are plotted. For now, only the first column in the condition file is taken into account. 
Expand to automatically plot all conditions. Ideally like subplots B and C in: 
 https://journals.plos.org/ploscompbiol/article/file?type=supplementary&id=info:doi/10.1371/journal.pcbi.1007147.s005

see conversation in #283.

dweindl:
@erikadudki : Is this still an issue, or has that been addressed in #429?

erikadudki:
To my knowledge, this is still an issue.

PR #429, opened 2020-04-21
Visualization column XValue not mandatory anymore

### LINKED ISSUE #283: column 'xValues' in visualization file should be mandatory ?
For the visualization routine the column 'xValues' is still mandatory, since it defines how the data should be plotted (e.g. time series or dose response).
Is that okay for everyone? 

- If yes, we should change it in the documentation. 
- If not, we should then agree on a default case to plot, if xValues is not provided.

erikadudki:
Automatic inferring of the column XValue: 
- checks in measurement-file if the time is constant:
-> if constant plot dose response plot (for now, only the first condition appearing in the experimentalCondition table, TODO: automatically plot all conditions, see #428) 
-> if not constant -> XValue = 'time' -> time series is plotted
- new: if grouping by observables, each observable will be plotted in one subplot (plotIds are inferred automatically according to observables)

closes #283

An enhancement of this issue is documented in #428.


petab/visualize/plot_data_and_simulation.py
```diff
@@ -123,7 +123,51 @@ def plot_data_and_simulation(
 
     # import visualization specification, if file was specified
     if isinstance(vis_spec, str):
+        # <<<<<<< HEAD
```
yannikschaelte, line 126: can this be deleted? or uncommented ... either of the two ...

PR #430, opened 2020-04-21
add sorting of indices of dataframes for the correct sorting of x-values

### LINKED ISSUE #414: something wrong with the ordering xValues in dose response
Maybe a bit difficult to explain here, but i will try my best :smile: 
I am generating a dose response plot and the some of the values in the condition table are not numerically sorted. This has been previously addressed in #396 
if I naturally sort (numerically) increasingly the `simulationConditionId` without changing the condition file, I get the expected plot:
Screenshot from 2020-03-25 17-18-17

However, I noticed that if I do the same BUT sorting the `simulationConditionId` by, e.g., `model1_data1`,`model1_data10`,`model1_data2`,`model1_data2` .... (not numerically increasing, default sorting) and accordingly the full experimentalData table is also sorted, I obtain this:
Screenshot from 2020-03-25 17-18-39

erikadudki:
- indices of the dataframes `conditions` and `ms` have to coincide, the problem here was, that they weren't. So now, index sorting is performed before `conditions` and `ms` are zipped together and then afterwards sorted according to ascending conditions. 
closes #414

### DIFF
--- a/petab/visualize/plotting_config.py
+++ b/petab/visualize/plotting_config.py
@@ -97,16 +97,16 @@ def plot_lowlevel(plot_spec: pd.Series,
 
         # construct errorbar-plots: noise specified above
         else:
-            # sort index for the case that indices of conditions and measurements differ
-            # if indep_var='time', conditions is a numpy array,
-            # for indep_var=observable its a Series
+            # sort index for the case that indices of conditions and
+            # measurements differ if indep_var='time', conditions is a numpy
+            # array, for indep_var=observable its a Series
             if isinstance(conditions, np.ndarray):
                 conditions.sort()
             elif isinstance(conditions, pd.core.series.Series):
                 conditions.sort_index(inplace=True)
             else:
-                print('strange: conditions object is neither numpy nor '
-                      'series...')
+                raise ValueError('Strange: conditions object is neither numpy'
+                                 ' nor Series...')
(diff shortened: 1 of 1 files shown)


petab/visualize/plotting_config.py
```diff
@@ -97,6 +97,18 @@ def plot_lowlevel(plot_spec: pd.Series,
 
         # construct errorbar-plots: noise specified above
         else:
+            # sort index for the case that indices of conditions and ms differ
```
yannikschaelte, line 100:
```suggestion
            # sort index for the case that indices of conditions and measurements differ
```

petab/visualize/plotting_config.py
```diff
@@ -97,6 +97,18 @@ def plot_lowlevel(plot_spec: pd.Series,
...
+            # if indep_var='time', conditions is a numpy array,
+            # for indep_var=observable its a Series
+            if isinstance(conditions, np.ndarray):
+                conditions.sort()
+            elif isinstance(conditions, pd.core.series.Series):
+                conditions.sort_index(inplace=True)
+            else:
+                print('strange: conditions object is neither numpy nor '
```
yannikschaelte, line 108: should this raise an error, i.e. the case never occur? then `raise ValueError("Strange: ...")`

yannikschaelte:
ping @erikadudki

Issue #319, opened 2020-02-21
visualization, multiple yValues in one plot


FFroehlich:
Documentation isn't very explicit about whether this should only be a single value or whether this can be multiple values (and if so in which format they should be passed).

FFroehlich:
Okay I think I now finally understood how it is supposed to be used. However, I feel like I have to provide an excruciating amount of information - why do I have to multiple xValues/xScale/xLabel/plotName for every observable. I believe allowing `yValues` and `legendEntry` to be some arraylike definition would be much more convenient.

JanHasenauer:
I think it would be great to have some default values here to reduce the need for specifications.

Suggestion:

1. xScale: default = line
2. xLabel:
    - if xLabel is time, then default = time
    - if xLabel is not time, then default = parameter/stateID
3. xOffset = 0

PR #431, opened 2020-04-21
default value for the column x_label in vis_spec

plakrisenko:
discussed in #319


plakrisenko:
> looks fine. This is also somehow addressed in the pull request #429, where I set the XLABEL to the condition name from the condition table, but this can then later be discussed which version we prefer :-)

I will have a look at your pull-request then!

yannikschaelte:
ping @plakrisenko

dweindl:
> ping @plakrisenko

PR #443, opened 2020-06-08
Make float parsing from CSV round-trip

lcontento:
`pandas` uses by default a C parser for CSV files for performance reasons. The default behaviour of such parser when parsing floating point values from strings is different from Python's behaviour. While Python parsing can be round-tripped (e.g., `float(0.999) == 0.999`) this is not true for the C parser.
I believe most users nowadays expect floats to be parsed correctly, so I added a keyword argument to all instances of `pandas.read_csv` asking for the Python-like behaviour. The only downside is that this behaviour is slower (the C code actually calls Python to perform the parsing). This may be more relevant for the parameter table and less important for measurement tables, so it may also be acceptable to enable it only for some tables and not others.

### DIFF
--- a/petab/conditions.py
+++ b/petab/conditions.py
@@ -23,7 +23,8 @@ def get_condition_df(
         return condition_file
 
     if isinstance(condition_file, str):
-        condition_file = pd.read_csv(condition_file, sep='\t')
+        condition_file = pd.read_csv(condition_file, sep='\t',
+                                     float_precision='round_trip')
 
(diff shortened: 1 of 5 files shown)


yannikschaelte:
speed should be no issue. if tests do not run from fork, consider creating a pull request from this repo (I added you).

lcontento:
Since tests are not completing, I will open a new pull request from a branch in this repo as @yannikschaelte suggested.

PR #446, opened 2020-06-18
Lint: detect duplicated observable IDs

lcontento:
The linter function `check_observable_df` does not detect duplicated observable IDs. MWE is
```python
import petab
import pandas as pd

observable_df = pd.DataFrame(data={
    petab.OBSERVABLE_ID: ['obs1', 'obs1'],
    petab.OBSERVABLE_FORMULA: ['x1', 'x2'],
    petab.NOISE_FORMULA: ['sigma1', 'sigma2']
}).set_index(petab.OBSERVABLE_ID)

petab.lint.check_observable_df(observable_df)
```
Incidentally, this causes a problem in `check_measurement_df` too.

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -133,6 +133,9 @@ def check_measurement_df(df: pd.DataFrame,
         if OBSERVABLE_TRANSFORMATION in observable_df:
             # Check for positivity of measurements in case of
             #  log-transformation
+            assert_observable_id_is_unique(observable_df)
+            # If the above is not checked, in the following loop
+            # trafo may become a pandas Series
             for measurement, obs_id in zip(df[MEASUREMENT], df[OBSERVABLE_ID]):
                 trafo = observable_df.loc[obs_id, OBSERVABLE_TRANSFORMATION]
                 if measurement <= 0.0 and trafo in [LOG, LOG10]:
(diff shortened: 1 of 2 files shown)


petab/lint.py
```diff
@@ -621,6 +625,21 @@ def assert_noise_distributions_valid(observable_df: pd.DataFrame) -> None:
                     f"table: {distr}.")
 
 
+def assert_observable_id_is_unique(observable_df: pd.DataFrame) -> None:
```
yannikschaelte, line 628:
```suggestion
def assert_unique_observable_ids(observable_df: pd.DataFrame) -> None:
```

petab/lint.py
```diff
@@ -621,6 +625,21 @@ def assert_noise_distributions_valid(observable_df: pd.DataFrame) -> None:
...
+    Arguments:
+        observable_df: PEtab observable DataFrame
+
+    Raises:
+        AssertionError: in case of problems
+    """
+    if len(observable_df.index) != len(set(observable_df.index)):
+        raise AssertionError(
+            f"{OBSERVABLE_ID} column in observable table is not unique.")
+
```
yannikschaelte, line 641: may be good to also return the duplicate ids.

petab/lint.py
lcontento: Name is indeed better. But since I essentially copied `assert_parameter_id_is_unique`, maybe if I change the name for the new function I should change it for the old too.

yannikschaelte: right. yes, makes sense to change the old one too then.

petab/lint.py
lcontento: You mean to put the duplicated list inside the `AssertionError` message? It is certainly more informative. Again, if I do it here I should do it for `assert_parameter_id_is_unique` too I guess.

yannikschaelte: not crucial, but I guess that would be helpful.

PR #459, opened 2020-08-05
Allow URL as filenames for YAML files and SBML models (Closes #187)

### LINKED ISSUE #187: Add support for retrieving remote files referenced in yaml config
- [ ] Adapt file existence checks (e.g. in `petab.CompositeProblem.validate`)
- [ ] Retrieve SBML models
- [ ] allow urls for specyfing yaml files

For tables it should be fine, as recent pandas versions can handle that.

dweindl:


petab/sbml.py
```diff
@@ -406,3 +405,62 @@ def write_sbml(sbml_doc: libsbml.SBMLDocument, filename: str) -> None:
...
+    from pandas.io.common import get_filepath_or_buffer, is_url, is_file_like
+
+    if is_file_like(filepath_or_buffer) or is_url(filepath_or_buffer):
+        buffer = get_filepath_or_buffer(filepath_or_buffer, mode='r')[0]
+        if is_url(filepath_or_buffer):
+            buffer = ''.join(line.decode('utf-8') for line in buffer)
+        else:
+            buffer = ''.join(line for line in buffer)
+
+        # URL or already opened file, we will load the model from a string
```
yannikschaelte, line 429: how do you know the file is opened already?

tests/test_petab.py
```diff
@@ -539,3 +539,15 @@ def test_to_files(petab_problem):  # pylint: disable=W0621
...
             .all().all()
+
+
+def test_load_remote():
+    """Test loading remote files"""
+
+    yaml_url = "https://raw.githubusercontent.com/PEtab-dev/petab_test_suite" \
+               "/master/cases/0001/_0001.yaml"
+    petab_problem = petab.Problem.from_yaml(yaml_url)
+
```
yannikschaelte, line 550: can e.g. also check nr of data / parameters, ... to see it is actually what is expected

petab/sbml.py
dweindl: From `if is_file_like(filepath_or_buffer) or is_url(filepath_or_buffer):` above. the former checks exactly that.

tests/test_petab.py
dweindl: Can, but I wouldn't find it too helpful. What would be a realistic scenario that would be covered by that? Keep it simple? :)

petab/sbml.py
dweindl: (and the latter would return a generator yielding the remote file content)

PR #463, opened 2020-08-24
fix #462

### LINKED ISSUE #462: writer functions do not perform sanity checks + use misleading options for writing tsv
When having dedicated writer functions such as `petab.conditions.write_condition_df` etc, I would actually expect that they perform some kind of sanity check on the output. Having pretty bare wrappers around `pd.to_csv` doesn't seem to helpful, especially since the writers use `index=True`, which is not consistent with the providede spec and, since the readers don't use `index_col=0`, leaves an `Unnamed 0` column in the imported DataFrame.

FFroehlich:


FFroehlich:
Looks like this identified a lot of copy paste errors where somebody bamboozled petab into writing observable frames as measurement frames. No immediate interest on my end to clean that up. Would recommend reducing that code duplication though ...

petab/conditions.py
```diff
@@ -48,8 +48,9 @@ def write_condition_df(df: pd.DataFrame, filename: str) -> None:
         df: PEtab condition table
         filename: Destination file name
     """
+    lint.check_condition_df(df)
     with open(filename, 'w') as fh:
-        df.to_csv(fh, sep='\t', index=True)
+        df.to_csv(fh, sep='\t', index=False)
```
dweindl, line 53: Hm. Why would you want to change that? We want conditionId as index.

FFroehlich: then we need to do something like `index_col=0` in https://github.com/PEtab-dev/PEtab/blob/768eb5b8c145a5de2ee6e054e61120d140de4be8/petab/conditions.py#L25

At the moment the index will be read as "Unnamed" column and have no effect whatsoever.

dweindl: Ah, I see. Okay, I think we want to have `index_col=CONDITION_ID` there.

FFroehlich: > Ah, I see. Okay, I think we want to have `index_col=CONDITION_ID` there.

That would remove the column from the DataFrame though. And `index=True` would still have no effect.

dweindl: > That would remove the column from the DataFrame though

Yes, that is what is expected everywhere for this moment.

petab/conditions.py
```diff
@@ -24,20 +24,12 @@ def get_condition_df(
...
     if isinstance(condition_file, str):
         condition_file = pd.read_csv(condition_file, sep='\t',
-                                     float_precision='round_trip')
+                                     float_precision='round_trip',
+                                     index_col=CONDITION_ID)
 
     lint.assert_no_leading_trailing_whitespace(
         condition_file.columns.values, "condition")
 
-    if not isinstance(condition_file.index, pd.RangeIndex):
```
dweindl, line 32: I think it was requested by somebody (@yannikschaelte ?) to have this function set the index correctly if not already done. Feels a bit funny, so okay for me to drop.

dweindl:
@yannikschaelte : Any thoughts on validating or not during reading/writing?

For me okay to do by default, but I'd make it optional at least.

petab/conditions.py
FFroehlich: > Ah, I see. Okay, I think we want to have `index_col=CONDITION_ID` there.

???

yannikschaelte: @dweindl Nope, not aware of these lines even ...

yannikschaelte:
> @yannikschaelte : Any thoughts on validating or not during reading/writing?
> 
> For me okay to do by default, but I'd make it optional at least.

Concerns with enforcing it could be slow-down for many files? Making it optional sounds like a good compromise to me.

petab/conditions.py
FFroehlich: Ah, this becomes relevant though when DataFrames are directly passed to this function and not read from file. The question is just whether DataFrame shouldn't have been sanitized before.

dweindl: Right. In my opinion they should be in the proper shape and just allowed to be passed through for convenience. But there were other opinions, but don't remember by whom. Okay for me to change.

FFroehlich:
Would appreciate if those who wrote the broken tests that are now flagged would fix them.

dweindl:
> Would appreciate if those who wrote the broken tests that are now flagged would fix them.

They were not broken. That was by design. It was was never in the contract, that writers would validate the content... 

Anyways: I updated some things and defaulted to no validation for backwards compatibility. Does the current state still work for you?

FFroehlich:
> > Would appreciate if those who wrote the broken tests that are now flagged would fix them.
> 
> They were not broken. That was by design. It was was never in the contract, that writers would validate the content...

Okay then I don't get the design of the tests. For example here: https://github.com/PEtab-dev/PEtab/blob/768eb5b8c145a5de2ee6e054e61120d140de4be8/tests/test_measurements.py#L10 the test checks whether an **observables** table roundtrips through measurements.get_**measurement**_df and measurements.write_**measurement**_df. If we want to keep this functionality, `measurements.get_measurement_df` and `measurements.write_measurement_df` cannot possibly have any functionality that is specific to measurements. In that case I really don't see the purpose of those wrappers since they are destined to remain bare wrappers around `pd.DataFrame.to_csv`. If thats the intention I would argue that those functions should be made private as they are useful for the library to convert internal representation to the specified format, but are pretty useless for the user since he is likely going to specify the table according pandas representation of the specified format.

> Anyways: I updated some things and defaulted to no validation for backwards compatibility. Does the current state still work for you?

I mean if you disable validation by default you may as well remove it because it will most likely not be activated when it could be useful.

dweindl:
> > > Would appreciate if those who wrote the broken tests that are now flagged would fix them.
> > 
> > 
> > They were not broken. That was by design. It was was never in the contract, that writers would validate the content...
> 
> Okay then I don't get the design of the tests. For example here:
> 
> https://github.com/PEtab-dev/PEtab/blob/768eb5b8c145a5de2ee6e054e61120d140de4be8/tests/test_measurements.py#L10
> the test checks whether an **observables** table roundtrips through measurements.get_**measurement**_df and measurements.write_**measurement**_df. 

No, it doesn't. This is an measurement table, although with only a subset of (elsewhere required) columns. (Compare https://petab.readthedocs.io/en/latest/documentation_data_format.html#measurement-table)

> If we want to keep this functionality, `measurements.get_measurement_df` and `measurements.write_measurement_df` cannot possibly have any functionality that is specific to measurements. In that case I really don't see the purpose of those wrappers since they are destined to remain bare wrappers around `pd.DataFrame.to_csv`. If thats the intention I would argue that those functions should be made private as they are useful for the library to convert internal representation to the specified format, but are pretty useless for the user since he is likely going to specify the table according pandas representation of the specified format.

Yes, it's only a thin wrapper around the pandas functions. I like using them, as it makes it easy in a codebase to find where the respective tables are written. Furthermore, it takes care of TSV and setting the index. Furthermore, it's convenient as it accepts both filenames and dataframes, so one doesn't have to distinguish between those cases when writing functions that should accept either. If none of those are relevant to you, feel free to not use it. I see no harm in having those public, and I found them useful in my own applications.

> > Anyways: I updated some things and defaulted to no validation for backwards compatibility. Does the current state still work for you?
> 
> I mean if you disable validation by default you may as well remove it because it will most likely not be activated when it could be useful.

When it could be useful is best judged by the user. Taking the same approach here as e.g. in libsbml which let's you save incomplete models (which only have to be valid xml, but nothing more). Here you can write incomplete files if you want to, e.g. if the files are generated over multiple stages with intermediate outputs being written to disk. (Again you could of course use plain pandas or pickle, or whatever instead.). In my applications, I validate the full set of files at the point at which I expect them to make sense together, so I don't want any redundant intermediate checking.

When adding the validation option to read/write functions here, one can of course argue whether the default should be on or off. Personally, I don't see enough benefit to break backwards compatibility here. Other opinions (including arguments) welcome...

FFroehlich:
> > > > Would appreciate if those who wrote the broken tests that are now flagged would fix them.
> > > 
> > > 
> > > They were not broken. That was by design. It was was never in the contract, that writers would validate the content...
> > 
> > 
> > Okay then I don't get the design of the tests. For example here:
> > https://github.com/PEtab-dev/PEtab/blob/768eb5b8c145a5de2ee6e054e61120d140de4be8/tests/test_measurements.py#L10
> > 
> > the test checks whether an **observables** table roundtrips through measurements.get_**measurement**_df and measurements.write_**measurement**_df.
> 
> No, it doesn't. This is a measurement table, although with only a subset of (elsewhere required) columns. (Compare https://petab.readthedocs.io/en/latest/documentation_data_format.html#measurement-table)

Okay, still not sure about the purpose of the test then. If the function doesn't write a measurement table, don't call it write_measurement_df. If it is supposed to write a partial measurement table, call it write_partial measurement_df.

> 
> > If we want to keep this functionality, `measurements.get_measurement_df` and `measurements.write_measurement_df` cannot possibly have any functionality that is specific to measurements. In that case I really don't see the purpose of those wrappers since they are destined to remain bare wrappers around `pd.DataFrame.to_csv`. If thats the intention I would argue that those functions should be made private as they are useful for the library to convert internal representation to the specified format, but are pretty useless for the user since he is likely going to specify the table according pandas representation of the specified format.
> 
> Yes, it's only a thin wrapper around the pandas functions. I like using them, as it makes it easy in a codebase to find where the respective tables are written. Furthermore, it takes care of TSV and setting the index. Furthermore, it's convenient as it accepts both filenames and dataframes, so one doesn't have to distinguish between those cases when writing functions that should accept either. If none of those are relevant to you, feel free to not use it. I see no harm in having those public, and I found them useful in my own applications.

Filename/Dataframe only applies to the readers, not the writers, which we were discussing here. Index setting was broken before this PR, so I don't see anybody benefitting from that before this PR.

> 
> > > Anyways: I updated some things and defaulted to no validation for backwards compatibility. Does the current state still work for you?
> > 
> > 
> > I mean if you disable validation by default you may as well remove it because it will most likely not be activated when it could be useful.
> 
> When it could be useful is best judged by the user. Taking the same approach here as e.g. in libsbml which let's you read/write incomplete models (which only have to be valid xml, but nothing more). Here you can write incomplete files if you want to, e.g. if the files are generated over multiple stages with intermediate outputs being written to disk. (Again you could of course use plain pandas or pickle, or whatever instead.). In my applications, I validate the full set of files at the point at which I expect them to make sense together, so I don't want any redundant intermediate checking.
> 
> When adding the validation option to read/write functions here, one can of course argue whether the default should be on or off. Personally, I don't see enough benefit to break backwards compatibility here. Other opinions (including arguments) welcome...

For intermediate results, I don't see the benefit of using these writers over `to_csv`. Still think that having functions that do what their name suggests would be beneficial over backwards compatibility. Fixing bugs, in a way, also breaks backwards compatibility. Closing for now though.

PR #472, opened 2020-10-06
Feature synthetic data

dilpath:
Ideas for tests? PEtab itself cannot simulate, just provides a base class to ensure compatibility when other simulation tools implement it.

The below is if you want to test implementation with AMICI/pyPESTO (there is no dependency on AMICI introduced with this PR).
Implementation in AMICI: [PR](https://github.com/AMICI-dev/AMICI/pull/1283)
Example optimisation with synthetic data: [notebook](https://github.com/ICB-DCM/pyPESTO/blob/feature_synthetic_data/doc/example/synthetic_data.ipynb), [PR](https://github.com/ICB-DCM/pyPESTO/pull/482)
PEtab base class: [code](https://github.com/PEtab-dev/PEtab/blob/feature_synthetic_data/petab/simulate.py), [PR](https://github.com/PEtab-dev/PEtab/pull/472)


petab/simulate.py
```diff
@@ -0,0 +1,154 @@
...
+        self.noise_formulas = petab.calculate.get_symbolic_noise_formulas(
+                self.petab_problem.observable_df)
+        self.rng = np.random.default_rng()
+
+    def clean_working_dir(self):
+        """
+        Deletes simulator-specific output files and their parent folder (see
+        the `__init__` method arguments).
+        """
+        shutil.rmtree(self.working_dir)
```
dweindl, line 45: This looks reeeeeally dangerous. Should at least be clearly documented in the constructor that this may be "cleaned" later on. Potentially issue a warning if it already exists / is non-empty?

dilpath: Thanks, good point, new logic is to delete the directory if it was created as a temporary directory, else only delete if `force=True`.

dilpath:
Added tests for noise and appropriately removing the working directory.

Also added a noise scaling factor.

petab/simulate.py
```diff
@@ -0,0 +1,188 @@
...
+import tempfile
+from typing import Dict, Optional, Union
+
+
+class Simulator(abc.ABC):
+    """
+    Base class that specific simulators should inherit.
+    Specific simulators should minimally implement the
+    `simulate_without_noise` method.
+    TODO link to amici.petab_simulate.PEtabSimulation as example
```
dweindl, line 17: to do :)

tests/test_simulate.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def simulate_without_noise(self) -> pd.DataFrame:
+        """Dummy simulation."""
+        return self.petab_problem.measurement_df
+
+
+@pytest.fixture
+def petab_problem() -> petab.Problem:
+    """Create a PEtab problem for use in tests."""
+    petab_yaml_path = 'doc/example/example_Fujita/Fujita.yaml'
```
dweindl, line 27: Would be good to use absolute paths to allow invoking pytest from any directory. (also below)

dweindl (approved):

PR #479, opened 2020-10-22
fix simulate

dilpath:
TODOs
- [X] default noise distribution wasn't handled in some cases
- [ ] ~~ensure synthetic data file rows are ordered as in the measurements file~~
  - should be done by the specific simulator (e.g. AMICI)
- [x] add option to ensure synthetic data is non-negative (can become negative after noise is added)


petab/simulate.py
```diff
@@ -122,6 +122,7 @@ def add_noise(
             self,
             simulation_df: pd.DataFrame,
             noise_scaling_factor: float = 1,
+            **kwargs
```
dweindl, line 125: Document where kwargs are forwarded to?

petab/simulate.py
```diff
@@ -143,6 +144,7 @@ def add_noise(
                 self.noise_formulas,
                 self.rng,
                 noise_scaling_factor,
+                **kwargs,
```
dweindl, line 147: Document where kwargs are forwarded to?

tests/test_simulate.py
```diff
@@ -65,6 +65,58 @@ def test_remove_working_dir(petab_problem):
...
+    # All measurements are non-zero
+    assert (synthetic_data_df['measurement'] != 0).all()
+    # No measurements are NaN
+    assert not (np.isnan(synthetic_data_df['measurement'])).any()
+
+    synthetic_data_df_with_noise = simulator.add_noise(
+        synthetic_data_df,
+    )
+    # Both negative and positive values are returned by default.
+    # This test will occasionally fail.
```
dweindl, line 97: Would be great if we can avoid that. Randomly failing tests waste so much time...

tests/test_simulate.py
```diff
@@ -65,6 +65,58 @@ def test_remove_working_dir(petab_problem):
...
+        (synthetic_data_df_with_noise['measurement'] >= 0).any(),
+    ])
+
+    synthetic_data_df_with_noise = simulator.add_noise(
+        synthetic_data_df,
+        zero_bounded=True,
+    )
+    # Values with noise that are different in sign to values without noise
+    # are zeroed.
+    # This test will occasionally fail.
```
dweindl, line 109: see previous

dweindl (approved):

petab/simulate.py
```diff
@@ -200,9 +221,20 @@ def sample_noise(
...
+    simulated_value_with_noise = getattr(rng, noise_distribution)(
         loc=simulated_value,
         scale=noise_value * noise_scaling_factor
     )
+
+    if (
+            zero_bounded and
+            np.sign(simulated_value) != np.sign(simulated_value_with_noise)
+    ):
+        return 0
```
dweindl, line 239:
```suggestion
        return 0.0
```

Since the function promises to return a `float`

dweindl (approved):
:+1:

PR #484, opened 2020-11-09
Doc: Clarify requirements for IDs

dweindl:
The general comment was repeatedly overlooked. Therefore, now added directly to the respective field descriptions.


LeonardSchmiester (approved):
:+1:

PR #487, opened 2020-11-10
Validator: output message in case of successful check is added

plakrisenko:


petab/petablint.py
```diff
@@ -162,6 +162,8 @@ def main():
             sys.exit(1)
 
     ret = petab.lint.lint_problem(problem)
+    if not ret:
+        print('PEtab format check completed successfully')
```
dweindl, line 166: When running `petablint` without `-v`, there is no output on purpose. I'd like to leave that as is.

If you are running `petablint -v`, you currently get an `OK` in case of no problems. If you think that is too uninformative, it's okay to put the message there (end of `petab.lint.lint_problem`).

dweindl (changes requested):

petab/petablint.py
jvanhoefer: looks good. That was, what I had in mind :)

petab/lint.py
```diff
@@ -744,7 +744,7 @@ def lint_problem(problem: 'petab.Problem') -> bool:
         logger.warning('Not all files of the PEtab problem definition could '
                        'be checked.')
     else:
-        logger.info('OK')
+        logger.info('PEtab format check completed successfully')
```
dweindl, line 747:
```suggestion
        logger.info('PEtab format check completed successfully.')
```

dweindl (approved):

PR #493, opened 2020-12-29
fix for pandas 1.2.0

dilpath:
Small change in pandas, `get_filepath_or_buffer` -> `_get_filepath_or_buffer` (not sure if `_get_filepath_or_buffer` should be used in PEtab since it's now named like a private method).

https://github.com/pandas-dev/pandas/pull/37639


yannikschaelte (approved):
I guess fine as a hot fix, but should be revised

setup.py
```diff
@@ -53,7 +53,7 @@ def absolute_links(txt):
       url='https://github.com/PEtab-dev/PEtab',
       packages=find_packages(exclude=['doc*', 'test*']),
       install_requires=['numpy>=1.15.1',
-                        'pandas>=1.0.1',
+                        'pandas>=1.2.0',
```
yannikschaelte, line 56: does get_handle not exist or behave differently for pandas < 1.2.0? but fine to enforce <=1.2.0 I think.

yannikschaelte (approved):
Nice! I suggest a new release then too as pyabc and pypesto are unhappy.

setup.py
dilpath: Does exist but may behave differently (the function that PEtab was using was "merged" with `get_handle` in the pandas PR that prompted this PR -- linked above). I haven't tested this PR with previous pandas versions.

yannikschaelte: thanks for the info, then enforcing :panda_face: >= 1.2.0 simply everywhere.

PR #511, opened 2021-03-02
Doc: Add editorial board 

dweindl:


fbergmann (changes requested):
I would prefer to add the images somewhere to doc/_static, just in case that we change institutions, or institute sites are not available for whatever reason. Also there we can already resize the images properly.

dweindl:
> I would prefer to add the images somewhere to doc/_static, just in case that we change institutions, or institute sites are not available for whatever reason. Also there we can already resize the images properly.

Works for me as well, if everybody is fine with having their photo persistently included here. Would be great if everybody could add one by themselves, or say whether they prefer use it from some other URL (@JanHasenauer @svenjakemmer @FFroehlich @fbergmann).

FFroehlich:
> > I would prefer to add the images somewhere to doc/_static, just in case that we change institutions, or institute sites are not available for whatever reason. Also there we can already resize the images properly.
> 
> Works for me as well, if everybody is fine with having their photo persistently included here. Would be great if everybody could add one by themselves, or say whether they prefer using one from some other URL (@JanHasenauer @svenjakemmer @FFroehlich @fbergmann).

For me the provided one is fine.

dweindl:
> > I would prefer to add the images somewhere to doc/_static, just in case that we change institutions, or institute sites are not available for whatever reason. Also there we can already resize the images properly.
> 
> Works for me as well, if everybody is fine with having their photo persistently included here. Would be great if everybody could add one by themselves, or say whether they prefer using one from some other URL (@JanHasenauer @svenjakemmer @FFroehlich @fbergmann).

Now handled via github profile pictures. Hope that's fine for everybody. Should be stable and addresses the resizing issue.

PR #521, opened 2021-03-25
Fix petab_scope_and_files.svg: 'estimated'->'estimate'

dweindl:
Parameter table in the figure had wrong column name. Reported by @dilpath.


dilpath (approved):
Looks good :+1:

The "ObservableTable" "noiseParameters" column entries aren't aligned, but fine to merge as is.

Issue #436, opened 2020-05-10
Add support for non-SBML models


dweindl:
Would be great if PEtab would be usable with non-SBML models. Formats to consider would include e.g. cellML, bngl, pysb, ....  Personally interested in pysb support. 

Will be a significant implementation effort, but probably worth it. Need to abstract from model implementation. For preparation, all `libsbml.Model` instances should be replaced by new `class Model`, abstract base class. Concrete implementations would be `SbmlModel`, `PysbModel`, ...

This does not imply any changes in file formats (yet). For the next format update, it would be good to already think about how to handle that e.g. in yaml files. Allow for `sbml_model_file`, `xx_model_file`, ... or add `model_type`?

Anybody interested in joining in for that?

dweindl:
Some notes from discussion with @FFroehlich about enabling use of PySB models:

#### PEtab problem definition: 

* Tables can be used as is, with minor limitations:
  * PySB models don't have meaningful species IDs and species names are a bit of a pain to be used as identifiers in any formulas. Therefore, observables and complex noise models would not be directly specified in `observableFormula` and `noiseFormula` in the measurement table, but one would reference `pysb.Expression` IDs there.
  * For the same reason, setting initial concentrations in the conditions table is problematic. This would be handled by having parameterized `pysb.Initial`s.
* [ ] YAML file entry `sbml_files` needs to be generalized or an additional field needs to be added. (Would one want to allow combining models of different formats?) 

#### PEtab library:

* Add model abstraction layer
  * [ ] Create `class Model` base class
  * [ ]  Gather all SBML-specific code in `class SbmlModel`
  * [ ] Create `class PysbModel`

#### Misc:

* [ ] Would be good to have a copy of the current PEtab test suite with PySB model instead of SBML model

paulflang:
I had the problem that my model was specified in BNGL and I wanted to create a PEtab problem from it to test several optimizers. SBML export from BNGL worked well, but PEtab did not support many of the characters used in the SBML export, so I wrote a script to remove them.

[...]

PR #538, opened 2022-03-18
Proposal: Different languages for model specification

dweindl:
# Motivation

There are a number of formats for specifying models in systems biology, each with their specific strengths and weaknesses. PEtab version 1.0.0 only allows  Systems Biology Markup Language (SBML) models. While SBML is supported by a large number of tools, there are good reasons to use other formats. For example, rule-based model formats (e.g., BioNetGenLanguage) permit more abstract and compact specification of models based on rules, which are generalisations of reactions. Therefore, and based on user request (#436), we propose to lift PEtab’s restriction to SBML models and allow arbitrary model formats.

# Proposed changes

* Changes to the PEtab YAML file:
  * Change `sbml_files` to `models`
  * `models` entries will be model IDs (following the existing conventions for PEtab IDs) mapping to:
    * `location`: path / URL to the model
    * `language`: model format
      Initial set of model format identifiers (to be extended as needed):
      * SBML: `sbml`
      * CellML: `cellml`
      * BNGL: `bngl`
      * PySB: `pysb`
  * An additional entry for mapping tables (see below) is added

  Example:

  **Before:**
  ```yaml
  format_version: 1
  parameter_file: parameters.tsv
  problems:
  - condition_files:
    - conditions.tsv
    measurement_files:
    - measurements.tsv
    observable_files:
    - observables.tsv
    sbml_files:
    - model1.xml
  ```

  **After:**
  ```yaml
  format_version: 2.0.0
  parameter_file: parameters.tsv
  problems:
  - condition_files:
    - conditions.tsv
    measurement_files:
    - measurements.tsv
    observable_files:
    - observables.tsv
    mapping_file: mappings.tsv # optional 
    models:
      id_for_model1:
        location: model1.xml
        language: sbml
  ```

* Changes to the format of existing tables/files:
  * Condition/Observable/Parameter Table
    All symbols that previously referenced the ID of SBML entities, such as parameter IDs or compartment IDs, now refer to (globally unique) named entities in the model, such as parameters, observables, expressions. For example, condition table columns may correspond to parameters, states, species of the referenced model. 
    For species, assignments in the condition table set the initial value at the beginning of the simulation for that condition, potentially replacing the initialization from preequilibration. For all other entities, values are statically replaced at all time points. For entities that assign values to other entities, such as SBML AssignmentRules, the value of the target of that rule is statically replaced at all time points.    
* Additional files
  * Mapping Table: 
    Mapping PEtab entity IDs to entity IDs in the model. This optional file may be used to reference model entities in PEtab files where the ID in the model would not be a valid identifier in PEtab (e.g., due to containing blanks, dots, or other special characters).
    The tsv file has two mandatory columns: `petabEntityId`, `modelEntityId`. Additional columns are allowed. modelEntityIds must be unique identifiers in the model. The mapping table must not map modelEntityIds to petabEntityIds that are also defined in any other part of the PEtab problem. modelEntityId may not refer to other petabEntityIds, including those defined in the mapping table. petabEntityIds defined in the mapping table may be referenced in condition, measurement, parameter and observable tables, but cannot be referenced in the model itself.
    For example, in SBML, local parameters may be referenced as `$reactionId.$localParameterId`, which are not valid PEtab IDs as they contain a `.` character. Similarly, this table may be used to reference specific species in a BGNL model which may contain many unsupported characters such as `,`, `(` or `.`. However, please note that IDs must exactly match the species names in the BNGL generated network file and no pattern matching will be performed. 

# Implications

* Tools need to check the model format and provide an informative message if the given format cannot be handled
* Validators will skip model-dependent validation when encountering unknown model types - ideally there would be some plugin mechanisms to provide validation

--- 

Co-authored by @FFroehlich @fbergmann. Also thanks to everybody participating in these discussions during the last COMBINE meeting.


paulflang:
> no pattern matching will be performed

So just that I understand this correctly, if I want to do pattern matching, I can still create a BNGL observable in the BNGL model like so
```
begin molecule types
    A(b~c~d)
end molecule types
begin observables
    Species myObservable A()
end observables
...
```
and use its ID `myObservable` in PEtab 2.0.0, for instance in the observable table (even without the mapping table, unless for whatever reason, I've decided to use an ID for the BNGL observable that contains characters that are not allowed in PEtab), but not in the condition table (as I would not know how that makes sense, also not if I map it to another `petabEntityId` in the mapping table, of course).

Anyway, awesome that PEtab will soon allow other model specifications. Thanks and congratulations!

FFroehlich:
> > no pattern matching will be performed
> 
> So just that I understand this correctly, if I want to do pattern matching, I can still create a BNGL observable in the BNGL model like so
> 
> ```
> begin molecule types
>     A(b~c~d)
> end molecule types
> begin observables
>     Species myObservable A()
> end observables
> ...
> ```
> 
> and use its ID `myObservable` in PEtab 2.0.0, for instance in the observable table (even without the mapping table, unless for whatever reason, I've decided to use an ID for the BNGL observable that contains characters that are not allowed in PEtab), but not in the condition table (as I would not know how that makes sense, also not if I map it to another `petabEntityId` in the mapping table, of course).

Yes, that is correct. In principle, you could also assign the value of the observable in the condition table, but there probably are only a few circumstances where this would make sense.

> 
> Anyway, awesome that PEtab will soon allow other model specifications. Thanks and congratulations!

Thank you!

doc/_static/petab_schema.yaml
```diff
@@ -38,13 +38,24 @@ properties:
...
-
-          items:
-            type: string
-            description: PEtab SBML file name or URL.
+        models:
+          type: object
+          description: One or multiple models
+
+          # the model ID
+          patternProperties: ^[a-zA-Z_]\w*$
```
dilpath, line 46: Could allow hyphens, but maybe problematic as the first character in YAML

doc/documentation_data_format.rst
```diff
@@ -41,12 +41,11 @@ Overview
...
 
 The PEtab data format specifies a parameter estimation problem using a number
-of text-based files (`Systems Biology Markup Language (SBML) <http://sbml.org>`_
-and
+of text-based files (
 `Tab-Separated Values (TSV) <https://www.iana.org/assignments/media-types/text/tab-separated-values>`_)
 (Figure 2), i.e.
 
-- An SBML model [SBML]
+- An model
```
dilpath, line 48:
```suggestion
- A model
```

dilpath: Could define currently supported formats here (e.g. the "Initial set of model format identifiers")

doc/documentation_data_format.rst
```diff
@@ -107,7 +109,7 @@ different experimental conditions).
 This is specified as a tab-separated value file in the following way:
 
 +--------------+------------------+------------------------------------+-----+---------------------------------------+
-| conditionId  | [conditionName]  | parameterOrSpeciesOrCompartmentId1 | ... | parameterOrSpeciesOrCompartmentId${n} |
+| conditionId  | [conditionName]  | modelEntityId1                     | ... | modelEntityId1${n}                    |
```
dilpath, line 112:
```suggestion
| conditionId  | [conditionName]  | modelEntityId1                     | ... | modelEntityId${n}                    |
```

doc/documentation_data_format.rst
```diff
@@ -140,12 +142,13 @@ Detailed field description
...
 
-- ``${parameterOrSpeciesOrCompartmentId1}``
+- ``${modelEntityId}``
 
-  Further columns may be global parameter IDs, IDs of species or compartments
-  as defined in the SBML model. Only one column is allowed per ID.
+  Further columns may be the IDs of model entities that have globally unique
+  IDs, such as parameters, species or compartments defined in the model.
+  Only one column is allowed per ID.
   Values for these condition parameters may be provided either as numeric
```
dilpath, line 150:
```suggestion
  Values for these condition entities may be provided either as numeric
```

doc/documentation_data_format.rst
```diff
@@ -140,12 +142,13 @@ Detailed field description
...
-  values, or as IDs defined in the SBML model, the parameter table or both.
+  values, or as IDs defined in the model, the parameter table or both.
```
dilpath, line 151: Could add mapping table.

doc/documentation_data_format.rst
```diff
@@ -157,14 +160,19 @@ Detailed field description
     concentration/amount of that species and will override the initial
     concentration/amount given in the SBML model or given by a preequilibration
     condition. If ``NaN`` is provided for a condition, the result of the
-    preequilibration (or initial concentration/amount from the SBML model, if
+    preequilibration (or initial concentration/amount from the model, if
     no preequilibration is defined) is used.
```
dilpath, line 164: Is an SBML assignment rule used if available, or would this have been removed from the model due to `float` values at other positions of the column for this condition entity?

doc/documentation_data_format.rst
```diff
@@ -679,6 +687,50 @@ Detailed field description
   legend and which defaults to the value in ``datasetId``.
 
 
+Mapping table
+-------------
+
+Mapping PEtab entity IDs to entity IDs in the model. This optional file may be
+used to reference model entities in PEtab files where the ID in the model would
+not be a valid identifier in PEtab (e.g., due to containing blanks, dots, or
```
dilpath, line 695:
```suggestion
not be a valid identifier in PEtab (e.g., due to inclusion of blanks, dots, or
```
Or can remove

doc/documentation_data_format.rst
```diff
@@ -679,6 +687,50 @@ Detailed field description
...
+other special characters).
+
+The tsv file has two mandatory columns, ``petabEntityId`` and
+``modelEntityId``. Additional columns are allowed.
```
dilpath, line 699:
```suggestion
The TSV file has two mandatory columns, ``petabEntityId`` and
``modelEntityId``. Additional columns are allowed.
```

doc/documentation_data_format.rst
```diff
@@ -679,6 +687,50 @@ Detailed field description
...
+
++---------------+---------------+
+| petabEntityId | modelEntityId |
++===============+===============+
+| STRING        | STRING        |
++---------------+---------------+
+| e.g.          | ...           |
```
dilpath, line 706:
```suggestion
```
Other tables do not include this `e.g.` row.

doc/documentation_data_format.rst
```diff
@@ -679,6 +687,50 @@ Detailed field description
...
+  itself.
+
+- ``modelEntityId`` [STRING, NOT NULL]
+
+  A globally unique identifier defined in the model.
+
+  For example, in SBML, local parameters may be referenced as
+  ``$reactionId.$localParameterId``, which are not valid PEtab IDs as they
+  contain a ``.`` character. Similarly, this table may be used to reference
+  specific species in a BGNL model which may contain many unsupported
```
dilpath, line 729:
```suggestion
  specific species in a BNGL model that may contain many unsupported
```

doc/documentation_data_format.rst
```diff
@@ -679,6 +687,50 @@ Detailed field description
...
+  characters such as ``,``, ``(`` or ``.``. However, please note that IDs must
+  exactly match the species names in the BNGL generated network file and no
```
dilpath, line 731:
```suggestion
  exactly match the species names in the BNGL-generated network file, and no
```

doc/documentation_data_format.rst
```diff
@@ -717,7 +769,7 @@ Parameter estimation problems combining multiple models
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Parameter estimation problems can comprise multiple models. For now, PEtab
-allows to specify multiple SBML models with corresponding condition and
+allows to specify multiple models with corresponding condition and
```
dilpath, line 772:
```suggestion
allows one to specify multiple models with corresponding condition and
```

doc/_static/petab_schema.yaml
```diff
@@ -38,13 +38,24 @@ properties:
...
       properties:
 
-        sbml_files:
-          type: array
-          description: List of PEtab SBML files.
-
-          items:
-            type: string
-            description: PEtab SBML file name or URL.
+        models:
```
dilpath, line 41: Could use `model_files`

dilpath (commented):
Just some suggestions, nothing major.

- "SBML" occurs a few times (e.g. in "doc/documentation_data_format.rst").
- The order of tables in "doc/documentation_data_format.rst" used to be alphabetical.

doc/_static/petab_schema.yaml
dweindl: Agreed. Would be more coherent.

doc/_static/petab_schema.yaml
dweindl: Could do that, or even allow for arbitrary strings. The current pattern is the same as for any other PEtab-related IDs, which would be easiest to remember.

dilpath: Ah, agreed :+1:

doc/documentation_data_format.rst
```diff
@@ -157,14 +161,19 @@ Detailed field description
...
-    preequilibration (or initial concentration/amount from the SBML model, if
+    preequilibration (or initial concentration/amount from the model, if
     no preequilibration is defined) is used.
 
   - ``${compartmentId}``
 
     If a compartment ID is provided, it is interpreted as the initial
     compartment size.
 
+  - For all other entities, values are statically replaced at all time points.
```
FFroehlich, line 172: In discussions with @dweindl, we have identified that this explanation is insufficient for rules that assign rates (where we only want to replace initial values not the rate at all timepoints). I am proposing the following text instead:

```
For all other entities, the value in the condition table either replaces the initial value or the value at all timepoints based on whether the model entity has a rate law assigned or not. For model entities that have constant algebraic assignments (but not necessarily constant values), i.e, that do not have a rate of change with respect to time assigned and that are not subject to event assignments, the algebraic assignment is replaced statically at all timepoints. Examples for such model entities are the targets of SBML `AssignmentRules`. 

For all other entities, e.g., those that are assigned by SBML `RateRules`, only the initial value value can be assigned in the condition table. If an assignment of the rate of change with respect to time or event assignment is desired, the values of model entities that are used to define rate of change or event assignments must be assigned in the condition table. If no such model entities exist, assignment is not possible.
```

This text may, in fact ,replace the definition above and parameters/compartments/species could be listed as examples of the two categories.

dweindl: Thanks @FFroehlich. Looks good to me. 

@fbergmann : What do you think?

fbergmann: @dweindl so it would mean, if the entity reference has an assignment rule, drop the assignment rule, right? That would essentially ensure that the value assigned holds for all timepoints. If that is what was meant, then its all good to me.

dweindl: Not really dropping the assignment rule, but rather replacing the `math` attribute of the `AssignmentRule` by the provided value. (Well, it's the same as dropping the `AssignmentRule` and replacing it by a different one with the same `variable`.)

dweindl: > Not really dropping the assignment rule, but rather replacing the value of the `math` attribute of the `AssignmentRule` by the provided value. (Well, it's the same as dropping the `AssignmentRule` and replacing it by a different one with the same `variable`.)

doc/documentation_data_format.rst
dweindl: The assignment rule should still be used in that case. Good candidate for a new test case.

doc/documentation_data_format.rst
dweindl: @FFroehlich I included your suggestion. Could you please have another look?

FFroehlich (approved):
👍

doc/_static/petab_schema.yaml
```diff
@@ -78,8 +91,12 @@ properties:
             type: string
             description: PEtab visualization file name or URL.
 
+        mapping_file:
```
dweindl, line 94: For all other file types we now allow for multiple files. It would be more coherent to also turn this into a list of filenames. Opinions?

FFroehlich: Agreed!

doc/documentation_data_format.rst
```diff
@@ -693,6 +704,48 @@ Detailed field description
...
+- ``petabEntityId`` [STRING, NOT NULL]
+
+  A valid PEtab identifier that is not defined in any other part of the PEtab
+  problem. This identifier may be referenced in condition, measurement,
+  parameter and observable tables, but cannot be referenced in the model
+  itself.
+
+- ``modelEntityId`` [STRING, NOT NULL]
+
+  A globally unique identifier defined in the model.
```
dweindl, line 739:
```suggestion
  A globally unique identifier defined in the model, *that is not a valid PEtab ID*.
```

That would make life much easier, since we won't have multiple PEtab parameter IDs mapping to the same model parameter. I don't see a good reason to support several aliases for the same entity, it will only make it more difficult to interpret the PEtab files.

CC @PEtab-dev/petab-editors

dilpath: Makes sense, but "valid PEtab ID" is a bit ambiguous to me. Maybe "that is not directly referenced in any other PEtab tables (its `petabEntityId` should be used instead)."

dweindl: > Maybe "that is not directly referenced in any other PEtab tables (its `petabEntityId` should be used instead)."

I am happy to change the wording, but my point was rather that it *cannot* be directly referenced elsewhere (due to identifier constraints), not only that it *is not* referenced elsewhere through it's original identifier.

dilpath: Ah right, then it makes sense :+1: The rest of the text about the mapping table makes this clear too. Looks good to me.

PR #537, opened 2022-03-12
Proposal for the introduction of extensions

FFroehlich:
This pull request describes a proposal for the extension of PEtab (see #528) that allows changes to the format through extensions. The proposed development and acceptance procedure is described in the changes to development.rst . The other change the documentation to indicate that the PEtab YAML file becomes mandatory.

Required steps for changes to the PEtab specification
-----------------------------------------------------
- [x] Updated format specifications (this PR)
- [x] Updated converter (no changes necessary)
- [x] Updated validator (no changes necessary)
- [x] Updated test suite (no changes necessary)
- [x] Updated changelog (this PR)

Note: this replaces #527, which was permanently closed when I accidentally deleted the corresponding repository

PR #545, opened 2022-05-24
Add `required` attribute to extensions in yaml file

### LINKED ISSUE #544: Extensions: required/optional
PEtab extensions were introduced in #537. We should be able to distinguish there between optional extensions and required extensions, i.e. those that modify the parameter estimation problem as such, and those that just add additional/optional information (e.g. annotations, info for visualization, ...). If some tool does not know about a certain optional extension, it can safely be ignored during import, if it does not know about a required extension, it should fail.

To include this information in the PEtab problem, I suggest to include a new field `required` in the yaml file, i.e. changing the schema to 
```yaml
  extensions:
    type: object
    description: |
      PEtab extensions being used.
    patternProperties:
      "^[a-zA-Z][\\-\\w]*$":

        type: object
        description: |
          Information on a specific extension
        properties:
          version:
            type: string
            pattern: ^([1-9][0-9]*!)?(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))*((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?$
          required:      # <------- NEW
            type: bool   # <------- NEW

        required:
          - version
          - required     # <------- NEW
      additionalProperties: true
```

Opinions?

dweindl:
PEtab extensions were introduced in #537. We should be able to distinguish there between optional extensions and required extensions, i.e. those that modify the parameter estimation problem as such, and those that just add additional/optional information (e.g. annotations, info for visualization, ...). If some tool does not know about a certain optional extension, it can safely be ignored during import, if it does not know about a required extension, it should fail.

This PR adds a `required` attribute to extensions in the yaml file to indicate whether they are required for the mathematical interpretation of the PEtab problem.

Resolves #544

### DIFF
--- a/doc/_static/petab_schema.yaml
+++ b/doc/_static/petab_schema.yaml
@@ -98,9 +98,14 @@ properties:
           version:
             type: string
             pattern: ^([1-9][0-9]*!)?(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))*((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?$
-
+          required:
+            type: bool
+            description: |
+              Indicates whether the extension is required for the
+              mathematical interpretation of problem.
         required:
           - version
+          - required
       additionalProperties: true
 
     additionalProperties: false


fbergmann (approved):
that looks good to me. thank you

FFroehlich (approved):
👍

PR #553, opened 2023-02-20
clarify amount/concentration

FFroehlich:


doc/documentation_data_format.rst
```diff
@@ -151,10 +151,11 @@ Detailed field description
   - ``${speciesId}``
 
     If a species ID is provided, it is interpreted as the initial
-    concentration/amount of that species and will override the initial
-    concentration/amount given in the SBML model or given by a preequilibration
+    condition of that species (as amount if hasOnlySubstanceUnits set to `True`
```
dweindl, line 154:
```suggestion
    condition of that species (as amount if `hasOnlySubstanceUnits` is set to `True` 
```

dweindl (approved):
Thanks :+1:

dweindl:
> @dweindl there is no test for this in the testsuite, right? 

Correct. Needs test. 

> Because current implementation in AMICI is wrong as far as I can tell.

:/

PR #558, opened 2023-05-11
Add PEtab.jl to README

sebapersson:
PEtab.jl is now a formal Julia package. Starting from the latest release 1.1 we test against the [PEtab test suite](https://github.com/sebapersson/PEtab.jl/blob/main/test/PEtab_test_suite.jl), and I am happy to announce (with some tweaking from PEtab.jl v1.0 -> v1.1) that we pass all tests. I have added this information to the README.


dweindl:
Cool, thanks!

PR #571, opened 2023-11-05
Explicitly exclude SBML local parameters from parameters table

### LINKED ISSUE #570: Global vs local SBML parameters
**Which problem would you like to address? Please describe.**
From my experience, PETab does not support SBML local parameters (for details on local vs global parameters, see for example p.53 of SBML 2 specification [here](https://raw.githubusercontent.com/combine-org/combine-specifications/main/specifications/files/sbml.level-2.version-5.release-1.pdf)).

**Describe the solution you would like**
It would be nice if the PEtab documentation could mention that all parameters must be declared globally.

dweindl:
Clarify that local parameters aren't allowed in the parameters table.

Closes #570

### DIFF
--- a/doc/documentation_data_format.rst
+++ b/doc/documentation_data_format.rst
@@ -443,6 +443,7 @@ and *must not* include:
   above)
 - Parameters included as column names in the *condition table*
 - Parameters that are AssignmentRule targets in the SBML model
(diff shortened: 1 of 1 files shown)


dilpath (approved):
:+1:

PR #579, opened 2024-05-16
Specification of math expressions

dweindl:
Previously, the math expression syntax wasn't specified. This was very problematic, because different libraries and programming languages have different names for the same functions, and more importantly, differ in operator precedence.

:eyes:  https://petab--579.org.readthedocs.build/en/579/documentation_data_format.html#math-expressions-syntax

Examples / test cases: https://github.com/dweindl/libpetab-python/blob/parse_math/tests/math/math_tests.yaml


doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+Literals
+~~~~~~~~
+
+Numbers
++++++++
+
+There is no distinction between integer and floating point numbers. Numbers
+can be specified in decimal notation, e.g. ``1.0``, ``-1.0``, ``1.0e-3``,
+``1.0e3``.
```
dilpath, line 798:
```suggestion
All numbers, including integers, are treated as floating point numbers. Numbers
can be specified in decimal and scientific notation, e.g. ``1.0``, ``-1.0``, ``1.0e-3``, ``1.0E3``. The general syntax in PCRE2 regex is ``\d*(\.\d+)?([eE][-+]?\d+)?``.
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+Booleans
+++++++++
+
+Boolean literals are ``true`` and ``false``. There is no automatic conversion
+between bool and float types. E.g., ``par * (par > 2)`` is not supported and
+should instead be implemented like ``piecewise(par, par > 2, 0)``.
```
dilpath, line 805:
```suggestion
Boolean literals are ``true`` and ``false``. There is no automatic conversion
between bool and float types. For example, ``par * (par > 2)`` is not supported, but can
be implemented with ``piecewise(par, par > 2, 0)``.
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+   "``()``", "1", "parentheses for grouping, acts like identity", "", "any single expression", "argument"
+   "``^``", "2", "exponentiation (shorthand for pow)", "right-to-left", "float, float", "float"
+   "``+`` ``-``", "3", "unary plus/minus", "right-to-left", "float", "float"
+   "``!``", "3", "not", "", "bool", "bool"
+   "``*`` ``/``", "4", "multiplication division", "left-to-right", "float, float", "float"
+   "``+`` ``-``", "5", "binary plus/minus, addition/subtraction", "left-to-right", "float, float", "float"
+   "``<`` ``<=`` ``>`` ``>=``", "6", "less than, less than or equal to, greater than, greater than or equal to", "left-to-right", "float, float", "bool"
+   "``==`` ``!=``", "6", "is equal to, is not equal to", "left-to-right", "float, float or bool, bool", "bool"
+   "``&&``, ``||``", "7", "logical and, logical or", "left-to-right", "bool, bool", "bool"
+   "``,``", "8", "comma  as function argument separator", "left-to-right", "any", ""
```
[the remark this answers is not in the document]
dilpath: Not sure if this will work, but there are RST tables that support newlines, and they might improve the look of these tables, e.g. less horizontal scrolling. If it works, I can do the same for the functions table.

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+    "``sin``, ``cos``, ``tan``, ``cot``, ``sec``, ``csc``", "float", "trigonometric functions", "float"
+    "``arcsin``, ``arccos``, ``arctan``, ``arccot``, ``arcsec``, ``arccsc``", "float", "inverse trigonometric functions", "float"
+    "``sinh``, ``cosh``, ``tanh``, ``coth``, ``sech``, ``csch``", "float", "hyperbolic functions", "float"
+    "``arcsinh``, ``arccosh``, ``arctanh``, ``arccoth``, ``arcsech``, ``arccsch``,", "float", "inverse hyperbolic functions", "float"
+    "``piecewise(true_value_1, condition_1, [true_value_2, condition_2, ] [...] [true_value_n, condition_n,] false_value)``", "``true_value*`` and ``false_value`` are either all float or all bool, ``condition*`` are all bool", "The function value is ``true_value_1`` if ``condition_1`` is true, else ``true_value_2`` if ``condition_2`` is true, else ..., else ``true_value_n`` if ``condition_n`` is true, else ``false_value``.", "float"
```
dilpath, line 823: Would try to make this display a little nicer -- let's see if the newlines above work first.

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+    "``abs(x)``", "float", "absolute value; ``piecewise(x, x>=0, -x)``", "float"
+    "``sign(x)``", "float", "sign of ``x``; ``piecewise(1, x>=0, -1)``", "float"
+    "``min(a, b)``, ``max(a, b)``", "float, float", "minimum / maximum of {a,b}; ``piecewise(a, a<=b, b)``, ``piecewise(a, a>=b, b)``", "float"
+
+Model time
+~~~~~~~~~~
+
+The model time is represented by the symbol ``time``. ``time`` always
+represents the current absolute simulation time (i.e., if the simulation starts
+at :math:`t_0 \neq 0`, then ``time`` is *not* the time since :math:`t_0`).
```
dilpath, line 833:
```suggestion
The model time is represented by the symbol ``time``, which is the current simulated time, not the current duration of simulated time; if the simulation starts
at :math:`t_0 \neq 0`, then ``time`` is *not* the time since :math:`t_0`.
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+Identifiers
+-----------
+
+* All identifiers in PEtab must consist only of upper and lower case letters,
+  digits and underscores, and must not start with a digit.
```
dilpath, line 839:
```suggestion
* All identifiers in PEtab may only contain upper and lower case letters,
  digits and underscores, and must not start with a digit. In PCRE2 regex, they must match `[a-zA-Z_][a-zA-Z_\d]*`.
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+* Identifiers are case-sensitive.
+
+* Identifiers must not be a reserved keyword (see below).
+
+* Identifiers must be globally unique within the PEtab Problem.
```
dilpath, line 845:
```suggestion
* Identifiers must be globally unique within the PEtab problem.
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+  Function names may be used as identifiers for other model entities, as they
+  can be distinguished by the context.
```
dilpath, line 847: Not sure this is nice for tool devs and I don't think there's much user benefit. Are any of our defined PEtab functions also commonly used as identifiers in models etc.?

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,118 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+Reserved keywords
+~~~~~~~~~~~~~~~~~
+
+The following keywords are reserved and must not be used as identifiers:
+
+* ``true``, ``false``: Boolean literals
+* ``inf``: Infinity
+* ``time``: Model time
+* ``nan``: reserved
```
dilpath, line 857:
```suggestion
* ``true``, ``false``: Boolean literals, used in PEtab expressions.
* ``inf``: Infinity, used in PEtab postequilibration measurements.
* ``-inf``: Negative infinity, used in PEtab timecourses with preequilibration.
* ``time``: Model time, used in PEtab expressions.
* ``nan``: Undefined in PEtab, but reserved to avoid implementation issues.
```

dilpath (approved):
:rocket:

doc/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: Also in favor of line breaks, but I thought I tried and failed with the csv tables. I'll test it again.

doc/documentation_data_format.rst
dweindl: Agreed, except for `-inf`. This is unary minus applied to `inf`, rather than a reserved keyword.

dilpath: Hm, IMO we currently treat it as a symbol, not as an expression. e.g. in the current PEtab timecourse proposal, we forbid expressions for the time points, hence the `-inf` is not an expression in a timecourse like `-inf:preeq_cond; 0:sim_cond`, so we have to call it a symbol, right? Or do we say `-inf` is the only allowable expression there? `-inf` also means "simulate from 0 until steady-state", which isn't incompatible with the "expression" interpretation of `-inf`, but feels more natural with the "symbol" interpretation of `-inf`. I guess no problem either way, fine to go with "expression" if we end up wanting to use it in expressions in a future extension.

dweindl: For the timecourses it's fine to treat it as a special symbol. If no expressions are allowed there, this whole section doesn't matter for the start time of the periods anyways. Using `-inf` as identifier for any other entity is anyways not allowed, since it's starting with a minus.

doc/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: Such a pain... but done.

doc/documentation_data_format.rst
dweindl: Changed. Better?

doc/src/functions.tsv
```diff
@@ -0,0 +1,13 @@
+Function	Argument types	Comment	Evaluates to
```
dweindl, line 1: Let's swap columns 2 and 3?

doc/src/functions.tsv
```diff
@@ -0,0 +1,13 @@
...
+``pow(a, b)``	float, float	power function `b`-th power of `a`	float
+``exp(x)``	float	exponential function pow(e, x);(`e` itself not a supported symbol,;but ``exp(1)`` can be used instead)	float
+``sqrt(x)``	float	square root of x;``pow(x, 0.5)``	float
+``log(a, b)``;``log(x)``;``ln(x)``;``log2(x)``;``log10(x)``	float[, float]	logarithm of ``a`` with base ``b``;``log(x, e)``;``log(x, e)``;``log(x, 2)``;``log(x, 10)``;(``log(0)`` is defined as ``-inf``);NOTE: ``log`` without explicit base is ``ln``, not ``log10``	float
+``sin``;``cos``;``tan``;``cot``;``sec``;``csc``	float	trigonometric functions	float
+``arcsin``;``arccos``;``arctan``;``arccot``;``arcsec``;``arccsc``	float	inverse trigonometric functions	float
+``sinh``;``cosh``;``tanh``;``coth``;``sech``;``csch``	float	hyperbolic functions	float
+``arcsinh``;``arccosh``;``arctanh``;``arccoth``;``arcsech``;``arccsch``,	float	inverse hyperbolic functions	float
```
dweindl, line 9:
```suggestion
``arcsinh``;``arccosh``;``arctanh``;``arccoth``;``arcsech``;``arccsch``	float	inverse hyperbolic functions	float
```

doc/src/functions.tsv
```diff
@@ -0,0 +1,13 @@
+Function	Argument types	Comment	Evaluates to
+``pow(a, b)``	float, float	power function `b`-th power of `a`	float
+``exp(x)``	float	exponential function pow(e, x);(`e` itself not a supported symbol,;but ``exp(1)`` can be used instead)	float
+``sqrt(x)``	float	square root of x;``pow(x, 0.5)``	float
```
dweindl, line 4:
```suggestion
``sqrt(x)``	float	square root of `x`;``pow(x, 0.5)``	float
```

doc/src/functions.tsv
```diff
@@ -0,0 +1,13 @@
...
+``sin``;``cos``;``tan``;``cot``;``sec``;``csc``	float	trigonometric functions	float
+``arcsin``;``arccos``;``arctan``;``arccot``;``arcsec``;``arccsc``	float	inverse trigonometric functions	float
+``sinh``;``cosh``;``tanh``;``coth``;``sech``;``csch``	float	hyperbolic functions	float
+``arcsinh``;``arccosh``;``arctanh``;``arccoth``;``arcsech``;``arccsch``,	float	inverse hyperbolic functions	float
+``piecewise(``;    ``true_value_1, condition_1,``;    ``[true_value_2, condition_2, ]``;    ``[...]``;    ``[true_value_n, condition_n,]``;    ``false_value``;``)``	``*value*``: all float or all bool;``condition*``: all bool	The function value is the ``true_value*`` for the first ``true`` ``condition``; or ``false_value`` if all conditions are ``false``.	float
+``abs(x)``	float	absolute value;``piecewise(x, x>=0, -x)``	float
+``sign(x)``	float	sign of ``x``;``piecewise(1, x>=0, -1)``	float
+``min(a, b)``;``max(a, b)``	float, float	minimum / maximum of {a,b};``piecewise(a, a<=b, b)``;``piecewise(a, a>=b, b)``	float
```
dweindl, line 13:
```suggestion
``min(a, b)``;``max(a, b)``	float, float	minimum / maximum of {`a`, `b`};``piecewise(a, a<=b, b)``;``piecewise(a, a>=b, b)``	float
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +741,321 @@ allows to specify multiple SBML models with corresponding condition and
...
+
+* Identifiers are case-sensitive.
+
+* Identifiers must not be a reserved keyword (see below).
+
+* Identifiers must be globally unique within the PEtab problem.
+  PEtab math function names must not be used as identifiers for other model
+  entities. Names of predefined function from the modeling framework are
+  allowed to be used for other model entities if supported by the framework,
+  but this is strongly discouraged.
```
dilpath, line 1048:
```suggestion
* Identifiers must be globally unique within the PEtab problem.
  PEtab math function names must not be used as identifiers for other model
  entities. PEtab does not put any further restrictions on the use of
  identifiers within the model, which means modelers could potentially
  use model-format--specific (e.g. SBML) function names as identifiers.
  However, this is strongly discouraged.
```

sebapersson (approved):
I am happy with all proposed changes, and think that the current syntax encapsulates the relevant functions.

As a potential addition, in the test-cases I just think three tests on Boolean <-> float conversion could be added (when everyone agree on that point).

fbergmann:
I'm happy with the proposed description. If anything I'd rename the `false_value`  in `piecewise` to `otherwise`. But that wont do much in the end.

PR #581, opened 2024-05-22
Specification of timecourses & long condition table

fbergmann:
This draft PR adds the descriptions from the [Time course specification](https://docs.google.com/document/d/1Uo0mtU8Tb6m0yZc-A-NprDXeS-9ZDYPA8U3Y_aChqEM/edit) to the `release/2.0.0` branch.

preview site is: <https://petab--581.org.readthedocs.build/en/581/v2/documentation_data_format.html>


doc/documentation_data_format.rst
```diff
@@ -166,6 +170,56 @@ Detailed field description
     If a compartment ID is provided, it is interpreted as the initial
     compartment size.
 
+  - `expressions`
+
+    Expressions containing more than a single parameter ID or numberical
```
dweindl, line 175:
```suggestion
    Expressions containing more than a single parameter ID or numerical
```

doc/documentation_data_format.rst
```diff
@@ -741,3 +791,71 @@ allows to specify multiple SBML models with corresponding condition and
 measurement tables, and one joint parameter table. This means that the parameter
 namespace is global. Therefore, parameters with the same ID in different models
 will be considered identical.
+
+
+Timecourses table
+-----------------
+
+The optional time courses tabke describes a sequence of different experimental
```
dweindl, line 799:
```suggestion
The optional time courses table describes a sequence of different experimental 
```

dweindl:
Resolved merge conflicts. 

This needs to be updated to match the new mostly-agreed-upon format in https://github.com/PEtab-dev/PEtab/issues/586 (changed experiments table, changed conditions table).

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
-+--------------+------------------+------------------------------------+-----+---------------------------------------+
 
+This is specified as a tab-separated value file in the following way:
+
+**TODO: keep conditionName or not?**
+
++--------------+------------------+------------------+-----------------+--------------------+
+| conditionId  | [conditionName]  | targetId         | operationType   | targetValue        |
++==============+==================+==================+=================+====================+
+| STRING       | [STRING]         | MODEL_ENTITY_ID  | STRING          | MATH_EXPRESSION    |
```
FFroehlich, line 160:
```suggestion
| PETAB_ID       | [STRING]         | NON_ESTIMATED_ENTITY_ID  | STRING          | MATH_EXPRESSION    |
```

doc/v2/documentation_data_format.rst
```diff
@@ -242,27 +354,25 @@ replicates and plot error bars.
 Detailed field description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- ``observableId`` [STRING, NOT NULL, REFERENCES(observables.observableID)]
+- ``observableId`` [STRING, REQUIRED, REFERENCES(observables.observableID)]
```
FFroehlich, line 357:
```suggestion
- ``observableId`` [PETAB_ID, REQUIRED, REFERENCES(observables.observableID)]
```

doc/v2/documentation_data_format.rst
```diff
@@ -205,13 +317,13 @@ model training or validation.
...
-| observableId | [preequilibrationConditionId] | simulationConditionId | measurement | time         |
-+==============+===============================+=======================+=============+==============+
-| observableId | [conditionId]                 | conditionId           | NUMERIC     | NUMERIC\|inf |
-+--------------+-------------------------------+-----------------------+-------------+--------------+
-| ...          | ...                           | ...                   | ...         | ...          |
-+--------------+-------------------------------+-----------------------+-------------+--------------+
++--------------+--------------+-------------+--------------+
+| observableId | experimentId | measurement | time         |
++==============+==============+=============+==============+
+| observableId | experimentID | NUMERIC     | NUMERIC\|inf |
```
FFroehlich, line 323:
```suggestion
| PETAB_ID | PETAB_ID | NUMERIC     | NUMERIC\|inf |
```

doc/v2/documentation_data_format.rst
```diff
@@ -242,27 +354,25 @@ replicates and plot error bars.
...
+- ``observableId`` [STRING, REQUIRED, REFERENCES(observables.observableID)]
 
   Observable ID as defined in the observable table described below.
 
-- ``preequilibrationConditionId`` [STRING OR NULL, REFERENCES(conditionsTable.conditionID), OPTIONAL]
-
-  The ``conditionId`` to be used for preequilibration. E.g. for drug
-  treatments, the model would be preequilibrated with the no-drug condition.
-  Empty for no preequilibration.
+- ``experimentId`` [STRING, REQUIRED, REFERENCES(experimentsTable.experimentID)]
```
FFroehlich, line 361:
```suggestion
- ``experimentId`` [[PETAB_ID], REQUIRED, REFERENCES(experimentsTable.experimentID)]
```

doc/v2/documentation_data_format.rst
```diff
@@ -242,27 +354,25 @@ replicates and plot error bars.
...
-- ``measurement`` [NUMERIC, NOT NULL]
+- ``measurement`` [NUMERIC, REQUIRED]
 
   The measured value in the same units/scale as the model output.
 
-- ``time`` [NUMERIC OR STRING, NOT NULL]
+- ``time`` [NUMERIC OR STRING, REQUIRED]
 
-  Time point of the measurement in the time unit specified in the SBML model, numeric value or ``inf`` (lower-case) for steady-state measurements.
+  Time point of the measurement in the time unit specified in the SBML model,
```
FFroehlich, line 374: Would consider it to be less confusing if ``inf`` was considered a numeric value here.

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+
+The optional experiments table describes a sequence of different experimental
+conditions (here: discrete changes) that are applied to the model.
+
+This is specified as a tab-separated value file in the following way:
+
++---------------------+-------------------+-----------------+
+| experimentId        | time              | conditionId     |
++=====================+===================+=================+
+| PETAB_ID            | NUMERIC or '-inf' | CONDITION_ID    |
```
FFroehlich, line 435: NUMERIC or '-inf', see above.

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+| intervention_effect |                40 | severe_lockdown |
++---------------------+-------------------+-----------------+
+
+Detailed field description
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The time courses table with three mandatory columns ``experimentId``,
+``time``, and ``conditionId``:
+
+- ``experimentId`` [STRING, REQUIRED]
```
FFroehlich, line 460:
```suggestion
- ``experimentId`` [PETAB_ID, REQUIRED]
```

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+
+  Identifier of the experiment. The usual PEtab identifier requirements apply.
+  This is referenced by the ``experimentId`` column in the measurement table.
+
+- ``time``: [NUMERIC or  ``-inf``, REQUIRED]
+
+  The time when the condition will become active, in the time unit specified
+  in the model. ``-inf`` indicates pre-equilibration (e.g., for drug
```
FFroehlich, line 468:
```suggestion
  in the model. ``-inf`` indicates that the respective condition will be simulated until a steady state is reached.
```

Strictly speaking equilibration would only be possible for models that satisfy detailed balance, would therefore prefer to avoid using (incorrect) jargon here.

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+  treatments, the model would be pre-equilibrated with the no-drug condition).
+
+- ``CONDITION_ID``: Reference to a condition ID in the condition table that
```
FFroehlich, line 471: [PETAB_ID, REQUIRED, REFERENCES(conditions.conditionID)]

doc/v2/documentation_data_format.rst
```diff
@@ -519,7 +699,7 @@ Additional columns may be added.
 Detailed field description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- ``parameterId`` [STRING, NOT NULL]
+- ``parameterId`` [STRING, REQUIRED]
```
FFroehlich, line 702:
```suggestion
- ``parameterId`` [PARAMETER_ID, REQUIRED]
```

doc/v2/documentation_data_format.rst
FFroehlich: **TODO**, behaviour if there is no steady state?

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+Experiments table
+-----------------
+
+The optional experiments table describes a sequence of different experimental
+conditions (here: discrete changes) that are applied to the model.
+
+This is specified as a tab-separated value file in the following way:
+
++---------------------+-------------------+-----------------+
+| experimentId        | time              | conditionId     |
```
FFroehlich, line 433: While I agree that we can drop the name in the conditions table, I think it would be nice to have an optional name column here.

doc/v2/documentation_data_format.rst
```diff
@@ -757,14 +939,14 @@ The TSV file has two mandatory columns, ``petabEntityId`` and
 Detailed field description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- ``petabEntityId`` [STRING, NOT NULL]
+- ``petabEntityId`` [STRING, REQUIRED]
```
FFroehlich, line 942:
```suggestion
- ``petabEntityId`` [PETAB_ID, REQUIRED]
```

FFroehlich (changes requested):

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
+    The target must be an algebraic target.
+
+  - ``noChange``: This value is used to indicate that no change is associated
+    with this condition. This is useful in cases where an explicit condition
+    ID is required, but no change is to be applied. (E.g., if pre-equilibration
+    is to be performed using the default values of the model, followed by a
+    different condition.)
+    ``targetId`` and ``targetValue`` have to be empty in this case.
+    For any conditionId with ``operationType=noChange``, there must not be any
+    other row  with the same ``conditionId`` in the conditions table.
```
dilpath, line 256:
```suggestion
    For any ``conditionId`` with ``operationType=noChange``, there must not be any
    other row with the same ``conditionId`` in the conditions table.
```

doc/v2/documentation_data_format.rst
dilpath: Agreed, in the measurements table it is `NUMERIC\|inf` instead

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+  is to be applied at the given `time`.
+
+  Note: The time interval in which a condition is applied includes the
+  respective starting timepoint, but excludes the starting timepoint of
+  the following condition. This means that for an experiment
+  ``[time_A:condition_A; time_B:condition_B]``, ``condition_A`` is active
+  during the interval ``[time_A, time_B)``. This implies that any event
+  assignment that triggers at ``time_B`` will occur *after* ``condition_B`` was
+  applied and for any measurements at ``time_B``, the observables will be
+  evaluated *after* ``condition_B`` was applied.
```
dilpath, line 481: This is the old (not long format) version of the experiments table, so it might be confusing to users

```suggestion
  Note: The time interval in which a condition is applied includes the
  respective starting timepoint, but excludes the starting timepoint of
  the following condition. This means that for an experiment where
  ``condition_A`` is applied at ``time_A`` and later ``condition_B`` is
  applied at ``time_B``, then ``condition_A`` is active
  during the interval ``[time_A, time_B)``. This implies that any event
  assignment that triggers at ``time_B`` will occur *after* ``condition_B`` was
  applied and for any measurements at ``time_B``, the observables will be
  evaluated *after* ``condition_B`` was applied.
```

doc/v2/documentation_data_format.rst
```diff
@@ -309,8 +419,77 @@ Detailed field description
...
+
+Multiple conditions can be applied at the same time point by specifying
+multiple lines with the same ``experimentId`` and ``time`` but different
+``conditionId``. The order of the conditions is arbitrary **????**.
+
+**TODO** how to deal with conflicts?
```
dilpath, line 487: Suggestion based on previous conversations, not sure how to word it best though
```suggestion
Multiple conditions can be applied at the same time point by specifying
multiple lines with the same ``experimentId`` and ``time`` but different
``conditionId``. The order of the conditions is arbitrary.

All changes across all simultaneous conditions are applied in order of
dependencies, with conservation of mass when volumes change, i.e.
a volume change implies a concentration change, but not vice versa.

Hence, volume changes are applied before concentration changes in the
same compartment, and changes to parameters are applied before
changes to species that depend on the parameters.
```

dilpath (commented):
Thanks, I'm happy to review this again after the current feedback round

dweindl:
Thanks for the feedback.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
+* Support for non-SBML models (:ref:`model`)
+* Changed condition table format (wide -> long) (:ref:`conditions_table`)
+* ``simulationConditionId`` and ``preequilibrationConditionId`` in the
+  :ref:`measurements_table` are replaced by ``experimentId`` and a more
+  flexible way for defining experiments / time courses
+  (:ref:`experiments_table`)
+* Support for math expressions in the condition table (:ref:`conditions_table`)
 * Clarification and specification of various previously underspecified aspects
-  (math expressions, overriding values in the condition table, etc.)
+  (:ref:`math_expressions`, overriding values in the condition table, etc.)
```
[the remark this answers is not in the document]
dweindl: To be updated once all changes here have been finalized.

doc/v2/documentation_data_format.rst
dweindl: "Finite numeric value or `inf`"? Since `-inf` would not be allowed?

doc/v2/documentation_data_format.rst
dweindl: Fine to add `experimentName` here if we drop the condition names. However, if we want to keep condition names, which then would have to be specified in some other place (yaml, annotations table, ...?), I would collect all of that in the same place.

doc/v2/documentation_data_format.rst
dweindl: Agreed. Any ideas for a more concise but still correct term?

In case of no steady state: `inf` objective value?

Do we want to specify any steady state criteria? We don't do that for integration accuracy and so on, but it feels problematic to not say anything here. We could suggest something like `dx/dt < 100 * integration_tolerance`?

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
+| conditionId1 | conditionName1   | modelEntityId2   |                 | 0.42               |
++--------------+------------------+------------------+-----------------+--------------------+
+| conditionId2 | ...              | modelEntityId1   |                 | modelEntityId1 + 3 |
++--------------+------------------+------------------+-----------------+--------------------+
+| conditionId2 | ...              | someSpecies      | setCurrentValue | 8                  |
++--------------+------------------+------------------+-----------------+--------------------+
+| ...          | ...              | ...              |                 | ...                |
++--------------+------------------+------------------+-----------------+--------------------+
+
+Each line specifies a change of one specific property of a specific entity.
```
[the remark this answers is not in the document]
dweindl: This hides the fact that a single entity has different types of associated values (e.g., current value vs. rate) and excludes "symbolic assignment" like `setAssignment`, doesn't it?

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
-    change or event assignments must be assigned in the condition table.
-    If no such model entities exist, assignment is not possible.
+- ``targetId`` [NON_ESTIMATED_ENTITY_ID, required except for ``operationType=noChange``]
+
+  The ID of the an entity that is to be changed when applying this condition.
+  This may an entity defined in the model, or an alias from the mapping
+  table. The entity must be uniquely identifiable. Parameters from the
+  :ref:`parameters_table` are not allowed here.
+  Different restrictions apply depending on the ``operationType`` and the
+  type of the model.
```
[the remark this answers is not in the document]
dweindl: > > This may an entity defined in the model, or an alias from the mapping
> > table.
> 
> `NON_ESTIMATED_ENTITY_ID` is already defined, so could move this statement about the mapping table to that definition.

:+1: 

> > The entity must be uniquely identifiable.
> 
> Do you have an example for why it's useful to state this here? Local parameters of SBML reactions might be? Still this is part of the definition of `NON_ESTIMATED_ENTITY_ID` I think, so could be moved there.

Can be moved there. Right, local parameters would be an example.

> > Parameters from the :ref:`parameters_table` are not allowed here.
> 
> Why can't I change the value of a non-estimated parameter that is in the parameter table? Fine for me, but now this also contradicts the definition of `NON_ESTIMATED_ENTITY_ID`, which explicitly supports parameter table parameters.

That wasn't allowed so far, and I would find it very confusing. Unlike SBML-parameters, PEtab parameters were always assumed to be constant throughout time and conditions. I would strongly prefer keeping it that way. Right, currently this is contradictory and `NON_ESTIMATED_ENTITY_ID` needs to be updated.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,213 @@ problem as such.
...
-| conditionId1 | [conditionName1] | 0.42                               | ... | parameterId                           |
-+--------------+------------------+------------------------------------+-----+---------------------------------------+
-| conditionId2 | ...              | ...                                | ... | ...                                   |
-+--------------+------------------+------------------------------------+-----+---------------------------------------+
-| ...          | ...              | ...                                | ... | ..                                    |
-+--------------+------------------+------------------------------------+-----+---------------------------------------+
 
+This is specified as a tab-separated value file in the following way:
+
+**TODO: keep conditionName or not?**
```
sebapersson, line 155: conditionName I think can be kept as optional as it can help with nicer plotting.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,216 @@ problem as such.
...
+  - ``setCurrentValue``: The current value of the target is initialized to the
+    value of ``targetValue`` when the respective condition is applied.
+    Any symbols in ``targetValue`` are interpreted as their value at the time
+    just before this change is applied (similar
+    to an SBML event with ``useValuesFromTriggerTime=True``; in case of the
+    first period of a timecourse, they will be interpreted as their initial
+    values as defined in the model).
+    The target must be a constant target or a differential target.
+
+  - ``setRate``: The time-derivative of the target is set to ``targetValue``
```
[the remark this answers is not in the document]
FFroehlich: Yeah, that sounds like it would be helpful to introduce `targetValue` in favour of having `addTo`, `multiply`, etc.

Why would we need operators like `rateOf`, `rateExpressionOf`? I don't see why one would ever transfer information from rate to value and everything else should be achievable one way or another.

Also, what about using `setValue` in favor of `setCurrentValue`?

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dilpath: > That wasn't allowed so far, and I would find it very confusing. Unlike SBML-parameters, PEtab parameters were always assumed to be constant throughout time and conditions. I would strongly prefer keeping it that way. Right, currently this is contradictory and `NON_ESTIMATED_ENTITY_ID` needs to be updated.

Yes, you're right, makes sense :+1:

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,213 @@ problem as such.
...
+
+    - Given a condition: ``condition1``: {``par1=par3``, ``par2=2*par3``}
+
+      This is allowed.
+
+    Condition changes should be implemented to respect the dependency
+    graph between model components:
+
+    - When a condition changes quantity ``A`` and ``B``, and ``B`` is dependent on
+      ``A``, the change in quantity A should be applied first such that the
```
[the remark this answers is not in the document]
dilpath: > This is not how SBML work right?
> 
> If I remember correctly if both `A` and `B` are assigned something in an event, basically the pre-event `A` and `B` values are used unless priority is specified.

I'm not sure, all I see in the SBML spec is (regarding simultaneously-occurring events)
> A given simulator may use any algorithm to choose an order as long as every event is executed exactly once.

This choice of compartment priority (and more generally to respect the dependency graph) is partly based on discussions with @matthiaskoenig last year -- perhaps @matthiaskoenig can say more here?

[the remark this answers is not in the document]
FFroehlich: There is a difference between simultaneously occuring events and simultaneous assignments during the same event. I would treat assignments via conditions the same as a single event, i.e. not subject to priority or useValuesFromTriggerTime attributes. I would propose to not follow the dependency graph (and rather simply use pre-assignment values) with the exception of compartments. As done with SBML Event Assignments (even though this isn't crystal clear in the sbml spec).

For example it might make sense to average two species A = B = (A + B)/2. Due to circularity, this assignment is not possible when following the dependency graph, but straightforward with pre-change values. Following of dependencies can always be reproduces through substitution.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,213 @@ problem as such.
...
+    (symbolically; the expression is not evaluated at the time the condition is
+    applied).
+    The target must be a differential target.
+
+  - ``addToRate``: The target's rate is increased by ``targetValue``
+    (symbolically).
+    The target must be a differential target.
+
+  - ``setAssignment``: The target is set to the value of ``targetValue``
+    (symbolically).
```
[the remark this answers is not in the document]
FFroehlich: No, this is discussed after Differential/Algebraic/Constant targets are introduced. To my understanding, structural alterations are not allowed.

doc/v2/documentation_data_format.rst
FFroehlich: In that case I would propose to move all names (including observables etc) to such a table. Would anyways be nice as it would introduce the possibility of, e.g. naming states.

matthiaskoenig:
In the timecourses, what we essentially have are event assignments triggered at t=t_trigger​. All event assignments must be executed simultaneously—this is not a step-by-step process. Instead, the model state must transition to the new state all at once. Any mathematical expressions that depend on updated parameters, compartments, or species must be re-evaluated accordingly. In other words, all changes are applied in a single step, and the AST (Abstract Syntax Tree) affected by these changes is fully re-evaluated. There is no predefined execution order—it is a single transformation of the model state.

Therefore, the claim that "compartment has priority" is incorrect and does not make sense. The underlying issue is that PETab changes are not clearly defined and have been incorrectly implemented by simulators. As seen in cases where changes affect both compartment volumes and concentrations, many simulators apply these changes sequentially, which is incorrect. Instead, all triggered changes must be applied in a single step. This clarification should also be explicitly stated for changes applied outside of timecourses.

Additionally, in SBML, there is a flag called useValuesFromTriggerTime, which allows the use of values before the assignments. This is likely what you are referring to:

> The time of assignment is not affected by the Event attribute useValuesFromTriggerTime—that attribute affects the time at which the EventAssignment’s math expression is evaluated. In other words, SBML allows decoupling the time at which the variable is assigned from the time at which its value expression is calculated.

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
matthiaskoenig: We want 2 different behaviors. These are managed in SBML events via the ` useValuesFromTriggerTime` flag. So we need probably this flag on the timecourse to ensure both scenarios are possible (see also comment below)

matthiaskoenig:
In SBML, all ListOf constructs are unordered lists. This means an SBML model can be modified by shuffling any ListOf elements, such as ListOfEventAssignments, without affecting the numerical results of a simulation. In other words, there is no inherent order in these lists, nor are there execution priorities—except in cases where ordering is necessary, in which attributes like priority have been introduced to dictate the execution of multiple Events.

The same principle should apply when multiple changes are made in PETab: all changes should be applied simultaneously. If ordering is required, an explicit order or priority attribute must be introduced. If mathematical expressions need to reference values from the end of the previous time span, a useValuesFromTriggerTime mechanism is necessary. For cases where a mix of behaviors is needed, an Event-like construct could be introduced, allowing assignments to be grouped and prioritized.

There is a reason SBML Events are complex. Either we adopt a full event-based approach where changes can be triggered after each time span, or we adhere to a simpler default approach where all changes are applied simultaneously.

One concern is that specifications often end up reflecting existing implementations rather than defining a general, correct solution. This appears to be happening here, as most PETab implementations currently apply changes sequentially. This approach is likely incorrect, but since the specification does not clearly define the intended behavior, there is uncertainty. Moreover, it does not align with what users of SBML models would expect.

doc/v2/documentation_data_format.rst
dweindl: The problem is that now there may be several rows per condition, which makes it not so ideal to specify any condition names there. Would you still keep it that table? Would all rows for that condition have to have the same value, or should there only be one non-empty entry per condition?

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: We might have different understandings of "structural alterations". Changing some algebraic expression `e := x` to `e := y` would be allowed, but changing some differential state to some algebraic expression would not be allowed. In how much that requires building different models is probably very much implementation dependent.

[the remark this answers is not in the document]
sebapersson: > We might have different understandings of "structural alterations". Changing some algebraic expression e := x to e := y would be allowed, but changing some differential state to some algebraic expression would not be allowed. In how much that requires building different models is probably very much implementation dependent.

Thanks for clarification, just a last question to fully understand. So if we have a parameter given via an assignment rule; `k = k1 * k2` would it then be allowed to change `k` via `setAssignment`?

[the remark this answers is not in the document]
dweindl: Yes, that was to my understanding the (only) use case for `setAssignment`.

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: One use case where one might want to use both the current value and the rate expression could be some piecewise definition of a rate expression where the condition depends on the current state. I'd don't consider super important.

dweindl:
> since the specification does not clearly define the intended behavior, there is uncertainty

Right. That definitely needs to be addressed.

> what users of SBML models would expect

To be honest, I am not sure for how many SBML users the users' expectations match the SBML standard when it comes to event assignments. I am still regularly confused when it comes to simultaneous changes in compartment sizes and concentrations.
That was one reason why considered different options, such as the compartment-change-first approach.
I'd find some explicit priority the most intuitive and least error-prone. This would in my opinion also cause the least issues with non-SBML models, where the interpretation or dependent/independent quantity is maybe less clear.

matthiaskoenig:
The main issue with establishing a fixed order for updating multiple variables (such as compartments, species, or parameters) is that mathematical calculations dependent on these variables could produce unexpected results depending on the update sequence.

For example, consider a compartment V and a concentration C. Any calculation that relies on both V and C may yield different outcomes depending on the update order. Suppose we start with V=1 and C=1, then update both to V=10 and C=10. If there is a calculation dependent on the relationship between V and C, the result will vary based on the order in which they are updated. For instance, an event triggered when V>C will only occur if V is updated before C. If C is updated first or both are updated simultaneously, the event may not trigger, leading to inconsistencies.

In other words, updating variables sequentially in a fixed order can lead to intermediate steps that place the model in an undefined state. This issue persists regardless of the specific order. Therefore, it is essential to either update all variables synchronously or clearly define and follow a predetermined update sequence for each variable to maintain a consistent model state (simply updating compartments before species does not resolve the issue).

The issue with the concentrations is just one of such a problem because species concentrations depend on their compartment volumes, so there is a math dependency linking C and V which depends on the order. But there can be many such dependencies in the math of the model.

dweindl:
Okay, fair point. The intermediate model states would be problem.

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: I'd vote for @FFroehlich's suggestion here.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +98,206 @@ problem as such.
...
+    **TODO** update
+
+  - ``noChange``: This value is used to indicate that no change is associated
+    with this condition. This is useful in cases where an explicit condition
+    ID is required, but no change is to be applied. (E.g., if pre-simulation
+    up to a steady state is to be performed using the default values of the
+    model, followed by a different condition.)
+    ``targetId`` and ``targetValue`` have to be empty in this case.
+    For any ``conditionId`` with ``operationType=noChange``, there must not be
+    any other row with the same ``conditionId`` in the conditions table.
```
dweindl, line 238: @FFroehlich suggested to completely remove `operationType` because after the recent simplification.  

`noChange` would then be specified by an empty conditionID in the experiments table.

Works for me.

@PEtab-dev/petab-editors : What are your thoughts?

doc/v2/documentation_data_format.rst
```diff
@@ -468,10 +460,15 @@ The time courses table with three mandatory columns ``experimentId``,
   returning ``NaN`` or ``inf`` values for the objective function.
   PEtab does not specify a numerical criterion for steady states.
 
-- ``CONDITION_ID``: [PETAB_ID, REQUIRED, REFERENCES(conditions.conditionID)]
+  For each ``experimentId`` the ``time`` values must be unique. The order of
+  the rows is arbitrary, but specifying the rows in ascending order of time
+  may improve human readability.
```
dilpath, line 465:
```suggestion
  Multiple conditions cannot be applied simultaneously in a single experiment,
  i.e. the ``time`` values must be unique for each ``experimentId``.
  The order of the rows is arbitrary, but specifying the rows in ascending order
  of time may improve human readability.
```

doc/v2/documentation_data_format.rst
dweindl: Shall we repurpose the mapping table for that? I.e. adding a `name` / `label` column there and allowing empty `modelEntityId`? This could also be a place for adding additional annotations, such as SBO terms.

Alternatively, adding that to the yaml file. I'd avoid another additional table.

doc/v2/documentation_data_format.rst
dweindl: To be continued at https://github.com/PEtab-dev/PEtab/pull/581/files#r1993877029; marking this as resolved.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,197 @@ problem as such.
...
+
+Any symbols in ``targetValue`` are interpreted as their value at the time
+just before this change is applied (similar
+to an SBML event with ``useValuesFromTriggerTime=True``; in case of the
+first period of a timecourse, they will be interpreted as their initial
+values as defined in the model).
+
+All changes are applied simultaneously,
+with conservation of mass when volumes change, i.e.
+a volume change implies a concentration change, but not vice versa.
```
FFroehlich, line 250:
```suggestion
All changes defined in the condition table are applied in two consecutive phases, similar to events in SBML.

1. **Evaluation of ``targetValues``**  
   ``targetValues`` are first evaluated using the *current*values of all variables in the respective expression.  
   - For the initial time period of an experiment, *current* values are determined by the initial conditions defined in the model.  
   - For subsequent time periods, the *current* values are defined by the simulation results at the end of the preceding time period.  

2. **Assignment and Update of Derived Variables**  
   Once evaluated, all ``targetValues`` are simultaneously assigned to their corresponding ``targetId``s, and derived variables are updated accordingly.  
   - This means that values of variables assigned by SBML assignment rules or PYSB expressions will need to be re-evaluated.  
   - Notably, for SBML models — where concentrations are always treated as derived variables — species with ``hasOnlySubstanceUnits=false`` and simultaneous changes to concentration and compartment volume in the condition table will have their amount and compartment volume updated first, with the resulting concentration computed as the fraction of amount to volume. For further details, refer to SBML semantic test suite case  
`01779 <https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/01779/01779-model.m>`_.
```

dweindl: Thanks, looks good. I'd only add some explanation similar to the one provided in the test case directly here. I'll commit the suggestion to make it easier for others to propose changes if necessary.

dilpath:
The suggestion for the reinitialization looks good to me. I could separate step 2 into "petab changes" and "model changes", if you agree. Here's a suggestion, that also includes some other details that we discussed.

```md
All changes defined in the condition table are applied in two consecutive phases, similar to events in SBML. The changes take precedence over other changes that may be encoded in the model, which occur in a third phase.

1. **Evaluation of ``targetValues``**
[...]

2. **Assignment of ``targetValues`` to ``targetIds``**  
All evaluated ``targetValues`` are simultaneously assigned to their corresponding ``targetId``s. It is invalid to apply more than one assignment to a single ``targetId`` simultaneously.

3. **Update of Derived Variables, Events and Finalization**
[...]
All other changes encoded in the model are now applied as well. For example, this includes SBML events. The model state for this timepoint is the model state after all of these changes, at the end of this phase.
```

dweindl:
> separate step 2

Thanks, I incorporated your suggestion with some modifications.

doc/v2/documentation_data_format.rst
```diff
@@ -298,19 +386,110 @@ Detailed field description
...
 
 
-Observable table
-----------------
+.. _v2_experiments_table:
+
+Experiments table
+-----------------
+
+The optional experiments table describes a sequence of different experimental
```
matthiaskoenig, line 405: One aspect I’m unsure about is how simple conditions (not part of a time course) are handled. I assumed that the Experiments table is required and that simple conditions would be treated via a time course with time = 0.

The Measurements table specifies that it requires experimentId ([PETAB_ID], REQUIRED, REFERENCES(experimentsTable.experimentID)), meaning it maps to experiments rather than conditions. So there would be a clarification how the simple cases of only conditions at time = 0 would be handled.

I would strongly prefer that the Experiments table is required. For an optimization problem one would need measurements which are linked to experiments and experiments require conditions. So both tables, conditions and experiments, are required. There can be the special case that both tables are empty in case of just mapping to the model as is without any changes.

doc/v2/documentation_data_format.rst
matthiaskoenig: Personally, I don't have an issue with information being duplicated in long tables. I.e. having the same id and name multiple times. This is probably easier to understand then to moving things to the mapping table.
If we start doing that I would prefer a solution which handles more metadata then just names for ids. E.g. annotations to ontologies, arbitrary information fields, ...
E.g. a nice hashmap which allows to annotate identifiers: 
```
{
    id: "condition1",  # id, table required
    table: "condition",
    name: "My condition"  # additional optional fields
    description: "This condition does ..."
    conditionClass: "class 1 - dosing",
    ...
}
```
For me the mapping table feels like a bad hack and prefer the duplication, or a more generic solution which allows to store additional information which can be used by tools.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+  assignment, i.e., they are not subject to time-derivative information, but
+  are not generally constant, e.g., the targets of SBML assignment rules.
 
+* **Constant entities**: Entities are that are defined in terms of a constant
+  value but may be subject to event assignments, e.g., parameters of an SBML
+  model that are not targets of rate rules or assignment rules.
+
+.. _v2_conditions_table:
+
+Conditions table
```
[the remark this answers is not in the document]
dweindl: I'd prefer introducing the independent tables before they are referenced by other tables (conditions, experiments, observables, measurements, parameters).

doc/v2/documentation_data_format.rst
dweindl: So far, the experiment table is required, unless a simulation is to be performed with the default model settings.

If there is only a single condition that is described by the model, experimentId for the respective measurements may be empty and the experiment table as well as the conditions table may be omitted.

I'd find it somewhat annoying to create those empty files in that case, but I don't feel very strongly about it.

Other opinions?

doc/v2/gfx/petab_scope_and_files.png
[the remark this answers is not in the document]
dweindl: I prefer introducing the conditionIds before they are used in the experiments table.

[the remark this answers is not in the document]
matthiaskoenig: no strong feeling about this one

doc/v2/documentation_data_format.rst
matthiaskoenig: I would prefer to have the experiments table required and have some way to fill out the conditions and experiments also for the special case of no changes. I.e. Having a row in the experiments table where the conditionId is empty (and an empty condition table). This would allow to use the value of the experimentId (experimentName) in plotting and analysis, ... . It is just a special case where the experiment has no changes.

dweindl: This option already exists, independent of requiring the experiment table. I.e. you can have a row in the experiment table with empty conditionId.

Also some experimentName could be provided there, or elsewhere (to be discussed, see other thread / mapping table).

doc/v2/documentation_data_format.rst
dweindl: > If we start doing that I would prefer a solution which handles more metadata then just names for ids. E.g. annotations to ontologies, arbitrary information fields, ...

This would be possible that way. The mapping table can have arbitrary additional columns (`description`, `sboTerm`, whatever you like). I am just not sure if we should worry about those here, as long as there is no defined use of this information. The same is true for `name`, just that I would include that in the specs, since there was a way to specify that in PEtab v1.

> mapping table feels like a bad hack

I'd also prefer not having it, but I don't have a better alternative. It's a compromise we have to make to make this table-based definition work, while keeping the number of tables manageable and avoiding duplication.

doc/v2/gfx/petab_scope_and_files.png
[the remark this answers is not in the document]
dweindl: Ok, bottom-up / top-down is a matter of taste. I'd leave it as is for now. This figure will have be updated a couple more times for PEtab v2 anyways. My main goal here was not displaying any incorrect information. I am fine with a complete makeover of that figure if somebody wants to take the time.

doc/v2/documentation_data_format.rst
matthiaskoenig: Sounds good to me. Then I am satisfied with the mapping table approach. 

We should not worry about these here. It would just be important that works in principle. I.e. use the mapping table to store additional information mapped on the ids which tools could use (mainly for visualization and communication).

doc/v2/documentation_data_format.rst
```diff
@@ -738,41 +920,51 @@ Detailed field description
...
 
-- ``modelEntityId`` [STRING, NOT NULL]
+  The ``petabEntityId`` may be the same as the ``modelEntityId``, but it must
+  not be used to alias an entity that already has a valid PEtab identifier.
+  This restriction is to avoid unnecessary complexity in the PEtab problem
+  files.
 
-  A globally unique identifier defined in the model,
-  *that is not a valid PEtab ID* (see :ref:`identifiers`).
+- ``modelEntityId`` [STRING, REQUIRED]
```
FFroehlich, line 963:
```suggestion
- ``modelEntityId`` [STRING or empty, REQUIRED]
```

doc/v2/documentation_data_format.rst
```diff
@@ -782,6 +974,11 @@ Detailed field description
   exactly match the species names in the BNGL-generated network file, and no
   pattern matching will be performed.
 
+- ``name`` [STRING, OPTIONAL]
```
FFroehlich, line 977:
```suggestion
- ``name`` [STRING or empty, OPTIONAL]
```

doc/v2/documentation_data_format.rst
```diff
@@ -782,6 +974,11 @@ Detailed field description
...
+
+  A human-readable name for the entity. This is optional and may be used for
+  reporting or visualization purposes.
```
FFroehlich, line 980: Say that if no name is provided it defaults to the petabEntityId?

FFroehlich (approved):

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+  value but may be subject to event assignments, e.g., parameters of an SBML
+  model that are not targets of rate rules or assignment rules.
+
+.. _v2_conditions_table:
+
+Conditions table
+----------------
+
+The conditions table specifies changes to parameters, initial or intermediate
+values of species, and compartments, during specific time periods.
```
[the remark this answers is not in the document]
dweindl: I find it a bit difficult. I already find the term "condition" a bit problematic, which basically describes a set of discrete changes. A "period" is the time between applying some condition and applying the next condition, or the end of the simulation. They aren't really interchangeable.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+     condition table will have their amount and compartment size updated
+     first, with the resulting concentration computed as the fraction of
+     amount to compartment size. For further details, refer to SBML semantic
+     test suite case
+     `01779 <https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/01779/01779-model.m>`_.
+
+3. **Update of Derived Variables**
+
+   After the assignment of the target values, all derived variables are
+   updated. If the model format has a concept of species and compartments,
```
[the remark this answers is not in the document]
dweindl: It can be combined with the previous step, but I think it's not the same. The one thing is updating the basic model state, and the other one, re-evaluating any dependent constructs. With concentration-species, this is distinction is a bit blurred, though.

I wouldn't say it's wrong for PySB models. If there are no concentrations to update, then this is simple skipped.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+
+   All other state- or time-dependent changes encoded in the model are now
+   applied as well. For example, this includes SBML events.
+   The model state for the evaluation of any trigger functions for this
+   timepoint is the model state after (3).
+
+   The initial model state for the new period at time = ``time``
+   (:ref:`v2_experiments_table`) the model state after all of these changes.
+
+5. **Evaluation of Observables**
```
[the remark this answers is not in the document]
dweindl: I found it more suitable here, but it's probably good to add a reference.

doc/v2/documentation_data_format.rst
```diff
@@ -298,19 +386,110 @@ Detailed field description
 
   The datasetId is used to group certain measurements to datasets. This is
   typically the case for data points which belong to the same observable,
-  the same simulation and preequilibration condition, the same noise model,
+  the same experiment, the same noise model,
```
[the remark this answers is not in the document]
dweindl: That sentence continues on the next line.

doc/v2/documentation_data_format.rst
```diff
@@ -298,19 +386,110 @@ Detailed field description
...
   the same observable transformation and the same observable parameters.
-  This grouping makes it possible to use the plotting routines which are
-  provided in the PEtab repository.
+  This grouping makes it possible to use the plotting routines that are
```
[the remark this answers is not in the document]
dweindl: The latter. I'd keep that for elsewhere. I don't think we should refer to the petab library here, but that's out of scope.

doc/v2/documentation_data_format.rst
```diff
@@ -298,19 +386,110 @@ Detailed field description
...
+  Note: The time interval in which a condition is applied includes the
+  respective starting timepoint, but excludes the starting timepoint of
+  the following condition. This means that for an experiment where
+  ``condition_A`` is applied at ``time_A`` and later ``condition_B`` is
+  applied at ``time_B``, then ``condition_A`` is active
+  during the interval ``[time_A, time_B)``. This implies that any event
+  assignment that triggers at ``time_B`` will occur *after* ``condition_B`` was
+  applied and for any measurements at ``time_B``, the observables will be
+  evaluated *after* ``condition_B`` was applied. For further details, refer to
+  :ref:`v2_reinitialization_semantics`.
```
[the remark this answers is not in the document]
dweindl: Agreed. Yes, good to clarify that.

doc/v2/documentation_data_format.rst
dweindl: Fine to add, but it's anyways not clear where exactly it would be used.

doc/v2/documentation_data_format.rst
```diff
@@ -47,18 +48,21 @@ Overview
...
+- (optional) An experiments file, which describes a sequence of different
+  experimental conditions that are applied to the model [TSV]
 
-- A parameter file specifying estimateable parameters and related information
+- An observables file specifying the observation model [TSV]
+
+- A parameters file specifying estimatable parameters and related information
   [TSV]
 
 - A grouping file that lists all of the files and provides additional information
```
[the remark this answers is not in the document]
dweindl: In terms of importance, yes. But it will be unclear which files we are referring to.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+:ref:`v2_experiments_table`, where multiple conditions can be combined to
+specify time courses or experiments that span multiple time periods.
+Any changes applied to the model will remain in effect until they are changed
+by a condition applied at a later time period or by the model itself.
+
+The conditions table is specified as a tab-separated value file in the
+following way:
+
++--------------+-------------------------+--------------------+
+| conditionId  | targetId                | targetValue        |
```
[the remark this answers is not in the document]
dweindl: Remove or add examples everywhere?

doc/v2/gfx/petab_files.png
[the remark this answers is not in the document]
dweindl: > Mappings should appear (I am not sure what this is and what it is good for?)

I left those out, because I guess that most users won't really need them, and it's hard to guess it's use from the figure.

> Probably the "Parameter transformations" should be removed, because as I remember these should be removed from the paramters table.

Correct, but that will be addressed in a separate PR. It's still in the spec for now.

Thanks for the suggestions. Let's update the rest once PEtab v2 is mostly complete.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+
+Conditions table
+----------------
+
+The conditions table specifies changes to parameters, initial or intermediate
+values of species, and compartments, during specific time periods.
+The start and end of the time period is defined in the
+:ref:`v2_experiments_table`, where multiple conditions can be combined to
+specify time courses or experiments that span multiple time periods.
+Any changes applied to the model will remain in effect until they are changed
```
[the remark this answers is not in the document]
FFroehlich: Probably best to drop it here, there is a more detailed description about model state/structure further below anyways.

doc/v2/documentation_data_format.rst
dweindl: @PEtab-dev/petab-editors / all: Optional or mandatory experiments table?

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
FFroehlich: Ah I see, will make a proposal how to split the description then. I think this will also make it more evident what's actually going on with SBML concentrations/volumes.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,188 @@ problem as such.
...
+     or PYSB expressions will need to be re-evaluated.
+
+   * Notably, for SBML models — where concentrations are always treated as
+     derived (algebraic) variables — species with ``hasOnlySubstanceUnits=false`` and
+     simultaneous changes to concentration and compartment size in the
+     condition table will have their amount and compartment size updated
+     first, with the resulting concentration computed as the fraction of
+     amount to compartment size. For further details, refer to SBML semantic
+     test suite case
+     `01779 <https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/01779/01779-model.m>`_.
```
FFroehlich, line 255:
```suggestion
   * These assignments respect language specific interpretations of what variables are *atomic* and which are *derived* or *algebraic*. Most notably, SBML considers species amounts and compartment size but **not** concentrations to be *atomic*. This means that for any entry that has a ``targetId`` that refers to a species with ``hasOnlySubstanceUnits=false`` will have it's ``targetValue`` (interpreted as concentration) converted to amounts using the **current** compartment size and then applied to the ``targetId``s amount. For further details, refer to SBML semantic
     test suite case `01779 <https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/01779/01779-model.m>`_.
```

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,188 @@ problem as such.
...
+
+3. **Update of Derived Variables**
+
+   After the assignment of the target values, all derived variables are
+   updated. If the model format has a concept of species and compartments,
+   this includes the update of species concentrations, which are derived from
+   their amount and compartment size.
```
FFroehlich, line 262:
```suggestion
   After the assignment of the target values, all derived variables, i.e. algebraic entities, are
   updated.
   
   * These updates respect language specific interpretations of which variables are *atomic* or *derived*/*algebraic*. For example, SBML considers concentrations to be *derived*, meaning that species concentrations will be recomputed based on (updated) values of species amounts and corresponding compartment sizes.
```

doc/v2/documentation_data_format.rst
FFroehlich: Would prefer making both tables optional over forcing the user to add empty dummies. Should be foolproof as the linter should catch missing experimentIds in the measurements table in case those tables were erroneously omitted.

doc/v2/documentation_data_format.rst
```diff
@@ -95,107 +100,189 @@ problem as such.
...
+  - ``STRING``: Any string
+  - ``NUMERIC``: Any number excluding ``NaN`` / ``inf`` / ``-inf``
+  - ``MATH_EXPRESSION``: A mathematical expression according to the
+    `PEtab math expression syntax <math_expressions>`_.
+  - ``PETAB_ID``: A string that is a valid PEtab ID
+  - ``NON_ESTIMATED_ENTITY_ID``: A string that is a valid PEtab ID and refers
+    directly, or via the mapping table, to an entity that has some associated
+    numeric value (e.g. a model parameter, parameter table parameter,
+    or a species in the model) and is not listed in the
+    :ref:`v2_parameters_table` (including ``estimate=0`` parameters).
```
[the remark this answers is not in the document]
FFroehlich: If we move this bit below the introduction of differential/constant entities we could simply say differential and constant entities that are not assigned in the parameters table, independent of the assigned value in the estimate column.

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: Tried to do that...

dweindl (approved):

matthiaskoenig:
Hi all,
I just realized that for steady state simulations, i.e. experiments with `time=-inf` I don't completely understand how this would work with models which have time-dependent events.

Currently it reads; "Any event triggers defined in the model must also be checked during this pre-simulation."
- it should probably be specified that the model should be simulated to steady state from `time=0`.
- it should be mentioned that models which are simulated to steady state should not have any time dependent triggers (as I understand it this is not well defined; and steady state solvers properly don't handle this correctly). Also if there are time dependent events one would want to start with time!=0 to pre-equilibration which is currently not possible.

So not clear to me how things work with "-Inf" time and time-triggered events.

dweindl:
> * it should probably be specified that the model should be simulated to steady state from `time=0`.

Agreed that it would be good to state that explicitly. I'd say from $t_0=0$, unless the model has specified some other $t_0$.


> * it should be mentioned that models which are simulated to steady state should not have any time dependent triggers (as I understand it this is not well defined; and steady state solvers properly don't handle this correctly). Also if there are time dependent events one would want to start with time!=0 to pre-equilibration which is currently not possible.

This is not only a problem with time-triggered events, but with any event assignments that are executed after the initial timepoint. If the model has such constructs, directly solving for $dx/dt = 0$ usually won't work. Not sure if this needs to be stated explicitly. If users don't want their events triggered during the steady state simulation, they can simple switch them off by including some indicator variable in the trigger function.

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+Conditions are applied at specific time points, which are defined in the
+:ref:`v2_experiments_table`. This allows for the specification of time
+courses or experiments spanning multiple time periods. A time period is the
+interval between two consecutive time points in the experiments table for a
+given experiment or the time between the last time point of an experiment and
+the time point of the last measurement for that experiment in the
+:ref:`v2_measurements_table`.
+
+The conditions tably only allows changes in the model state, not the model
+structure. That mean that only constant or differential entities (e.g.,
```
dweindl, line 184:
```suggestion
structure. That means that only constant or differential entities (e.g.,
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+  A unique identifier for the experiment, referenced by the ``experimentId``
+  column in the :ref:`v2_measurements_table`.
+
+- ``time``: [NUMERIC or ``-inf``, REQUIRED]
+
+  The time at which the specified condition will become active, in the time
+  unit specified in the model. ``-inf`` indicates that the respective
+  condition will be simulated until a steady state is reached.
+
+  If the simulation of a condition with steady fails to reach a steady state,
```
dweindl, line 360:
```suggestion
  If the simulation of a condition with steady state fails to reach a steady state,
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+  In such cases, PEtab interpreters should notify the user, for example, by
+  returning ``NaN`` or ``inf`` values for the objective function.
+  PEtab does not specify a numerical criterion for steady states.
+  Any event triggers defined in the model must also be checked during this
+  pre-simulation.
+
+  Multiple conditions cannot be applied simultaneously within in a single
+  experiment; each ``time`` values must be unique for each ``experimentId``.
+
+  ``time`` will override any initial time specified in the model.
```
dweindl, line 372:
```suggestion
  ``time`` will override any initial time specified in the model, except in the case of ``time`` =  `ìnf``, in which case the model-specified time will be used (or 0, if the model does not explicitly specify an initial time).
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+  If the simulation of a condition with steady fails to reach a steady state,
+  and the condition is required for the evaluation of simulation at
+  measurement points, the evaluation of the model is not well-defined.
+  In such cases, PEtab interpreters should notify the user, for example, by
+  returning ``NaN`` or ``inf`` values for the objective function.
+  PEtab does not specify a numerical criterion for steady states.
+  Any event triggers defined in the model must also be checked during this
+  pre-simulation.
+
+  Multiple conditions cannot be applied simultaneously within in a single
```
dweindl, line 369:
```suggestion
  Multiple conditions cannot be applied simultaneously within a single
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+  sequences of experimental conditions applied to the model [TSV].
 
-- A parameter file specifying estimateable parameters and related information
-  [TSV]
+- An :ref:`observables file <v2_observables_table>` defining the observation
+  model [TSV].
 
-- A grouping file that lists all of the files and provides additional information
-  including employed extensions [YAML]
+- A :ref:`parameters file <v2_parameters_table>` defining estimable
```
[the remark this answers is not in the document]
sebapersson: To note close the door on Bayesians :), I would write 

I think parameters to be estimated or inferred.

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
sebapersson: Second this that examples can help digest the standard.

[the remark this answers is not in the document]
dweindl: Agreed, although in the longer run, I'd prefer some self-contained problem-specific HowTos / examples outside this document. I think this will be more useful.

sebapersson (approved):
Looks good, feel free to merge for me when some of the minor comments below are fixed.

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,372 @@ This document explains the PEtab data format.
...
+  parameters and related information [TSV].
```
dweindl, line 72:
```suggestion
- A :ref:`parameters file <v2_parameters_table>` to set parameter values globally and to specify the parameters to be estimated as well as their parameter bounds and prior distributions [TSV].
```

matthiaskoenig (approved):

doc/v2/gfx/petab_scope_and_files.png
fbergmann, line 1: @dweindl  should the measurement table not change to list an experimentId now rather than a timecourseId?

doc/v2/gfx/petab_scope_and_files.svg
fbergmann, line 1: @dweindl same here

dweindl: Indeed. Thanks.

doc/v2/gfx/petab_scope_and_files.png
dweindl: fixed.

matthiaskoenig (commented):
The parameter value of the upper bound 250 in the parameter table is not a number:
25o -> 250

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
-compartments for specific simulation conditions (generally corresponding to
-different experimental conditions).
+* **Algebraic entities**: Entities that are defined in terms of an algebraic
+  assignment rather than time derivatives. They are not necessarily constant,
+  for example, the targets of SBML assignment rules.
 
-This is specified as a tab-separated value file in the following way:
+* **Constant entities**: Entities are that are defined in terms of a constant
+  value but may be subject to event assignments, e.g., parameters of an SBML
+  model that are not targets of rate rules or assignment rules.
```
dilpath, line 163: It's unclear that algebraic assignments like SBML initialAssignmentRules are constant entities. I'm not sure how to best word that here. In PEtab SciML there is an attempt at distinguishing these by calling this "prior to simulation". I guess the difference between algebraic entities and constant entities is that constant entities are piecewise constant (e.g. initialAssignmentRules or eventAssignments can affect them, but otherwise they are constant), and algebraic entities are everything else.

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+environment of the system of interest. These modifications are referred to as
+(experimental) *conditions*.
+
+Conditions are applied at specific time points, which are defined in the
+:ref:`v2_experiments_table`. This allows for the specification of time
+courses or experiments spanning multiple time periods. A time period is the
+interval between two consecutive time points in the experiments table for a
+given experiment or the time between the last time point of an experiment and
+the time point of the last measurement for that experiment in the
+:ref:`v2_measurements_table`.
```
dilpath, line 181: Since this is a user's choice, I don't think we need to specify that the simulation stops at the last measurement. A tool might want to give users an easier way to perform predictions, than providing a dummy measurements table.

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
-+--------------+------------------+------------------------------------+-----+---------------------------------------+
+.. _v2_conditions_table:
 
-Row- and column-ordering are arbitrary, although specifying ``conditionId``
-first may improve human readability.
+Conditions table
+----------------
 
-Additional columns are *not* allowed.
+The conditions table defines discrete changes to the model. These (sets of)
```
dilpath, line 170: To match the start of the Experiments table section

```suggestion
The optional conditions table defines discrete changes to the model. These (sets of)
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
-  (math expressions, overriding values in the condition table, etc.)
-* Support for extensions
-* Observable IDs are now allowed to be used in observable/noise formulas
+* Support for models in other formats than SBML (:ref:`v2_model`).
+* Revised condition table format from wide to long format
+  (:ref:`v2_conditions_table`).
+* ``simulationConditionId`` and ``preequilibrationConditionId`` in the
+  :ref:`v2_measurements_table` are replaced by ``experimentId`` and a more
+  flexible way for defining experiments and time courses
+  (:ref:`v2_experiments_table`).
```
dilpath, line 131: And also now optional, if the user wants to simulate the SBML model as-is

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+The experiments table is provided as a tab-separated values (TSV) file with
+the following structure:
+
++---------------------+-------------------+-----------------+
+| experimentId        | time              | conditionId     |
++=====================+===================+=================+
+| PETAB_ID            | NUMERIC or '-inf' | CONDITION_ID    |
++---------------------+-------------------+-----------------+
+| timecourse_1        |                 0 | condition_1     |
++---------------------+-------------------+-----------------+
```
dilpath, line 324: To match the conditions table section

```suggestion
+---------------------+-------------------+-----------------+
| experimentId        | time              | conditionId     |
+=====================+===================+=================+
| PETAB_ID            | NUMERIC or '-inf' | CONDITION_ID    |
+---------------------+-------------------+-----------------+
| e.g.                |                   |                 |
+---------------------+-------------------+-----------------+
| timecourse_1        |                 0 | condition_1     |
+---------------------+-------------------+-----------------+
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+| intervention_effect |                40 | severe_lockdown |
++---------------------+-------------------+-----------------+
+
+The order of the rows is arbitrary, but specifying the rows in ascending order
+of time may improve human readability.
+
+Detailed field description
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The time courses table has three mandatory columns ``experimentId``,
```
dilpath, line 346:
```suggestion
The experiment table has three mandatory columns ``experimentId``,
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+  and the condition is required for the evaluation of simulation at
+  measurement points, the evaluation of the model is not well-defined.
+  In such cases, PEtab interpreters should notify the user, for example, by
+  returning ``NaN`` or ``inf`` values for the objective function.
+  PEtab does not specify a numerical criterion for steady states.
+  Any event triggers defined in the model must also be checked during this
+  pre-simulation.
+
+  Multiple conditions cannot be applied simultaneously within a single
+  experiment; each ``time`` values must be unique for each ``experimentId``.
```
dilpath, line 370:
```suggestion
  Multiple conditions cannot be applied simultaneously within a single
  experiment; each ``time`` value must be unique for each ``experimentId``.
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+
+  ``time`` will override any initial time specified in the model,
+  except in the case of ``time`` =  `ìnf``, in which case the model-specified
+  time will be used (or 0, if the model does not explicitly specify an initial
+  time).
```
dilpath, line 375:
```suggestion
  ``time`` will override any initial time specified in the model,
  except in the case of ``time`` =  ``-ìnf``, in which case the model-specified
  time will be used (or 0, if the model does not explicitly specify an initial
  time).
```

doc/v2/documentation_data_format.rst
```diff
@@ -14,187 +15,375 @@ This document explains the PEtab data format.
...
+
+
+- ``CONDITION_ID``:
+  [PETAB_ID or empty, REQUIRED, REFERENCES(condition.conditionId)]
```
dilpath, line 379:
```suggestion
- ``conditionId``:
  [CONDITION_ID or empty, REQUIRED, REFERENCES(condition.conditionId)]
```

dilpath: Hm not sure if this is correct. I guess it should be "PETAB_ID", but the table above it says "CONDITION_ID".

doc/v2/documentation_data_format.rst
```diff
@@ -501,6 +711,8 @@ One row per parameter with arbitrary order of rows and columns:
 +=============+=================+=========================+=============+============+==============+==========+=====+
 | STRING      | [STRING]        | log10\|lin\|log         | NUMERIC     | NUMERIC    | NUMERIC      | 0\|1     | ... |
```
dilpath, line 712:
```suggestion
| PETAB_ID      | [STRING]        | log10\|lin\|log         | NUMERIC     | NUMERIC    | NUMERIC      | 0\|1     | ... |
```

doc/v2/documentation_data_format.rst
```diff
@@ -739,41 +954,67 @@ Detailed field description
...
-| petabEntityId | modelEntityId |
-+===============+===============+
-| STRING        | STRING        |
-+---------------+---------------+
-| reaction1_k1  | reaction1.k1  |
-+---------------+---------------+
++---------------+---------------+---------------+
+| petabEntityId | modelEntityId | [name]        |
++===============+===============+===============+
+| STRING        | STRING        | STRING        |
```
dilpath, line 992:
```suggestion
| PETAB_ID        | STRING        | STRING        |
```

doc/v2/documentation_data_format.rst
dweindl: Change to "between the last time point and the end of the simulation" ?

doc/v2/documentation_data_format.rst
FFroehlich: experiment*s*

doc/v2/documentation_data_format.rst
dilpath: The thing that is unclear to me is that both of these are algebraic, but one is compatible with "constant entity" whereas the other is not:
- initialAssignmentRule is an algebraic equation that is evaluated at one time point
- assignmentRule is an algebraic equation that is evaluated throughout time

dweindl: I tried to rephrase it, please check again.

doc/v2/documentation_data_format.rst
dweindl: Updated.

doc/v2/documentation_data_format.rst
dilpath: "Experiment table" in the figure, but fine for me

dweindl: I think that is inconsistent throughout the documentation, library, ...
Let's address that separately and get this monster merged and out of the way.

doc/v2/documentation_data_format.rst
dilpath: Looks good, the intention of the different entities is clear now :+1: Thanks

doc/v2/documentation_data_format.rst
matthiaskoenig: I could imagine adding an empty time interval at the end without any condition with the time>last timepoint. I.e. one could have a time interval at the end where no measurements exist. In this case the simulation has to run until the last specified time point of the experiment.
This should be clarified and would provide a simple manner to encode the end of the simulation if one wants, e.g. for visualization or other purposes.

dweindl: I'd consider that out of scope here. For the purpose of parameter estimation, tools can just stop at the last measurement timepoint.

Your use case would have to be addressed in some visualization function. For proper visualization, additional information such as the output timepoints would be required. I don't think we should try to get that in here.

matthiaskoenig: Or could I just add a dummy timepoint to the measurement table with a NaN value at the end?

dweindl: > Or could I just add a dummy timepoint to the measurement table with a NaN value at the end?

That would currently be considered an error. NaN is not allowed. We had that implemented at some point, but it's an abuse of the measurements table. We can continue that discussion outside this PR if you'd consider it important to change that.

If the only purpose is visualization, I'd rather specify the endpoint and potentially time-resolution in the visualization table (or any better replacement of that...). 

Let's get this thing here merged and revisit any revision of visualization-related functionality afterwards?

PR #584, opened 2024-07-02
Fix petab_files.svg

dweindl:
The old version wasn't properly editable.
Therefore, I recreated the figure with slight deviations.


dilpath (approved):
Looks great! Thanks a lot.

PR #590, opened 2024-12-04
Clarify meaning of `initializationPrior*`

### LINKED ISSUE #587: Clarify meaning of `initializationPrior*`
Initially, `initializationPriorType` and `initializationPriorParameters` were introduced to provide prior distributions for sampling initial points in a multi-start optimization setting. For other global optimization schemes it is less clear how this should be incorporated. Not at all? For the initial population? Whenever a new point is sampled? ...?

dweindl:
Consensus seems to be that the use of initialization priors is optional.

Closes #587.


dilpath (approved):
Fine for me :+1:

PR #595, opened 2025-02-17
Add new parameter prior distributions 

sebapersson:
I have added information for the new `initializationPriorType` and `objectivePriorType` distributions, along with support for truncated distributions in the PEtab v2 specification. Additionally, a mathematical definition is provided for each prior, since some distributions, such as the gamma distribution, can be parameterized in different ways.

I would keep this as a draft until it is decided whether `parameterScale` should remain part of the PEtab standard. If `parameterScale` is retained, then I believe we should also keep the `parameterScale` priors where it makes sense. This is because inference is sometimes performed directly on `log(p)` (the logarithm of parameter `p`), as the geometry of the posterior landscape is often more favorable, and to avoid issues with transforming prior distributions, users sometimes place a prior on the log of the parameter directly—even though a prior on the log scale can be nonintuitive (for example, a uniform prior on the log scale is not uniform on the linear scale :).

I will also update the test-suite to test for correct prior evaluation on `objectivePriorType` for the linear priors (including testing truncated priors).


doc/v2/documentation_data_format.rst
```diff
@@ -620,7 +642,69 @@ Detailed field description
 
   Prior parameters used for the objective function during estimation.
   For more detailed documentation, see ``initializationPriorParameters``.
+  As for initialization priors, truncated priors are allowed.
+
+Prior distributions
+~~~~~~~~~~~~~~~~~~~
 
+For ``initializationPriorParameters`` and ``objectivePriorType`` several
```
dilpath, line 650:
```suggestion
For ``initializationPriorType`` and ``objectivePriorType`` several
```

or did you mean `*PriorParameters` for both? In any case, I guess one needs to be changed.

doc/v2/documentation_data_format.rst
```diff
@@ -620,7 +642,69 @@ Detailed field description
...
+distributions are supported (see above). Denote the parameter value by :math:`x`
+and the Gamma function by :math:`\Gamma`; then, the following parameterizations
+are used for the supported priors:
+
+- *cauchy*:
+  - parameters: (location, scale) = :math:`(\mu, \sigma)`
```
dilpath, line 656: The formatting is off (currently these two lines are rendered as one line). Maybe a newline helps.
https://petab--595.org.readthedocs.build/en/595/v2/documentation_data_format.html#prior-distributions
```suggestion
- *cauchy*:

  - parameters: (location, scale) = :math:`(\mu, \sigma)`
```

dilpath (approved):
Happy to review again before merging.

> I would keep this as a draft until it is decided whether `parameterScale` should remain part of the PEtab standard

:+1: 

> a mathematical definition is provided for each prior

:+1:

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,61 @@ Detailed field description
 
 - ``initializationPriorType`` [STRING, OPTIONAL]
 
-  Prior types used for sampling of initial points for estimation. Sampled
-  points are clipped to lie inside the parameter boundaries specified by
-  ``lowerBound`` and ``upperBound``. Defaults to ``parameterScaleUniform``.
+  Prior types used for sampling initial points for estimation. If the prior is
+  not truncated (see ``initializationPriorParameters``), sampled points are
+  clipped to lie within the parameter boundaries specified by ``lowerBound``
```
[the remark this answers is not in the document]
sebapersson: I would be happy to avoid having the sampled points clipped, as I agree this is a bit weird and can for users lead to unexpected behavior. Furthermore, with truncated bounds can now be enforced in a more logical way.

doc/v2/documentation_data_format.rst
sebapersson: It worked when I compiled locally to a pdf, but alas. I think a newline with help, and will add it for all distributions.

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,61 @@ Detailed field description
...
+  and ``upperBound``. Otherwise, points are sampled directly from the truncated
+  distribution. The default is ``parameterScaleUniform``.
 
-  Possible prior types are:
+  Possible prior types are (for mathematical definition see below):
 
-    - *uniform*: flat prior on linear parameters
-    - *normal*: Gaussian prior on linear parameters
+    - *cauchy*, Cauchy prior on linear parameters
```
matthiaskoenig, line 588: Use consistent separation characters: sometimes ",", sometimes ":"

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,61 @@ Detailed field description
...
+    - *chisquare*, Chisquare prior on linear parameters
+    - *exponential*, Exponential prior on linear parameters
+    - *gamma*, Gamma prior on linear parameters
     - *laplace*: Laplace prior on linear parameters
     - *logNormal*: exponentiated Gaussian prior on linear parameters
-    - *logLaplace*: exponentiated Laplace prior on linear parameters
+    - *normal*: Gaussian prior on linear parameters
+    - *rayleigh*, Rayleigh prior on linear parameters
+    - *uniform*: Flat prior on linear parameters
+    - *parameterScaleCauchy*, Cauchy prior on original parameter scale
```
matthiaskoenig, line 597: I would add the `parameter*` to a new list and start it with something like:
Supported `parameterScales` are: 
- *parameterScaleLaplace*: Laplace prior on original parameter scale
...

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,61 @@ Detailed field description
...
+    - *cauchy*, Cauchy prior on linear parameters
+    - *chisquare*, Chisquare prior on linear parameters
+    - *exponential*, Exponential prior on linear parameters
+    - *gamma*, Gamma prior on linear parameters
     - *laplace*: Laplace prior on linear parameters
     - *logNormal*: exponentiated Gaussian prior on linear parameters
-    - *logLaplace*: exponentiated Laplace prior on linear parameters
+    - *normal*: Gaussian prior on linear parameters
+    - *rayleigh*, Rayleigh prior on linear parameters
+    - *uniform*: Flat prior on linear parameters
```
matthiaskoenig, line 596: As I understand things *uniform** is the default if nothing is provided. If yes, please add the `(default)` to uniform. If not ignore this comment

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,61 @@ Detailed field description
...
+
+  So far, only numeric values are supported, no parameter names. Parameters for
+  the different prior types are:
+
+    - *cauchy*, location; scale
+    - *chisquare*, degrees of freedom
+    - *exponential*, rate
+    - *gamma*, shape; scale
+    - *laplace*: location; scale
+    - *logNormal*: parameters of corresp. normal distribution (see: normal)
```
matthiaskoenig, line 627: use consistent separators, i.e. `,` or `:`

matthiaskoenig (commented):
Some minor comments

doc/v2/documentation_data_format.rst
sebapersson: It is the `parameterScaleUniform` which is default (should already be stated).

doc/v2/documentation_data_format.rst
sebapersson: Thanks for noticing!

dweindl (commented):
Great, thanks!

> I would keep this as a draft until it is decided whether parameterScale should remain part of the PEtab standard.

Agreed.

doc/v2/documentation_data_format.rst
```diff
@@ -575,41 +577,65 @@ Detailed field description
...
 
-    - *uniform*: flat prior on linear parameters
-    - *normal*: Gaussian prior on linear parameters
+    - *cauchy*: Cauchy prior on linear parameters
+    - *chisquare*: Chisquare prior on linear parameters
+    - *exponential*: Exponential prior on linear parameters
+    - *gamma*: Gamma prior on linear parameters
     - *laplace*: Laplace prior on linear parameters
     - *logNormal*: exponentiated Gaussian prior on linear parameters
-    - *logLaplace*: exponentiated Laplace prior on linear parameters
```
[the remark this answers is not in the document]
sebapersson: Thanks for the feedback.

I agree that objective priors should also be unbounded (unless a truncated prior is explicitly used). We can briefly discuss and reach a conclusion on this at the next editor meeting.

As for log-priors, I would opt for only supporting log-normal as it is quite a common distribution, and it appears sometimes in biology, e.g. it has been argued that for example protein distributions follow a log-normal. Moreover, most distribution packages support log-normal (and its truncated variant). The other log distributions I would drop, and suggest in the spec that one can parameterize the model if one for example needs a log-laplace. I will also raise this on the next editor meeting.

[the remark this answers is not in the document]
dweindl: Since we want to support lossless petab v1->v2 conversion, we probably need to include the previous log/log10 versions for normal/laplace in some way.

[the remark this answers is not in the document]
sebapersson: There is no `log10` version of the normal and Laplace priors in PEtab v1.

However, v1 -> v2 is a fair point. Then we should keep the log-laplace prior, and I will add it and its pdf to the documentation.

Should be noted that the tricky part with the log-laplce is that it does not appear in all distribution packages. But, since the Laplace distribution is standard in most libraries, generating samples from it should still be quite straightforward.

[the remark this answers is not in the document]
sebapersson: > There is, with parameterScale=log10 and objectivePriorType=laplace, isn't it?

This will be relevant to fix if decide to drop the `parameterScale` column in the parameter table.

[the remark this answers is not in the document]
dweindl: > if decide to drop the `parameterScale` column in the parameter table.

Shall we address that first then?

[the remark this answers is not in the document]
dweindl: At least on the editor side. I think we will have a final vote on the complete PEtab v2 in the end.

I think it might be better to remove it first from the current v2 draft, and stay focused on the actual prior changes here. (The experiments pull request took us almost a year... smaller bites might be faster).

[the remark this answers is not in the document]
sebapersson: I agree, I will just focus on the priors here and not do any other changes to the parameter table. FYI, I am currently waiting for the next editor meeting first to see what is decided regarding the naming of the prior column, and whether we should remove the `initializationPrior`.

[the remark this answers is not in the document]
sebapersson: > Let's create a PR for that as basis for the discussion?

So first we merge this with `initializationPrior`, `objectivePrior` and `parmeterScale` priors, then we create a separate PR with removing `initializationPrior` and renaming `objectivePrior`? If that is the case, sounds good for me to partition into smaller (hopefully easier to merge) PRs.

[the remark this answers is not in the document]
dweindl: I'd go the other way around: 

1a. drop parameter scale (also removes some of the prior options)
1b. drop intialization prior, rename objective prior
2. update what's left of the parameter priors, then we don't need to worry about initializationPrior and the interaction with parameter scale anymore in here 

1a/b can be done in parallel.

Either is fine though.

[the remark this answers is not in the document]
sebapersson: This makes more sense, and I agree we should do it this way.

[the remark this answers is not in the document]
matthiaskoenig: Please make this changes so we get rid of unnecessary things (we agreed on that in the last meeting and email discussion):
> 1a. drop parameter scale (also removes some of the prior options) (EDIT: https://github.com/PEtab-dev/PEtab/pull/611)
> 1b. drop intialization prior, rename objective prior

Then it is much easier to get the formulations right for the rest.

doc/v2/documentation_data_format.rst
```diff
@@ -795,46 +797,70 @@ Detailed field description
...
-    - logLaplace: parameters of corresp. Laplace distribution (see: laplace)
-    - parameterScaleUniform: lower bound; upper bound
-    - parameterScaleNormal: mean; standard deviation (**not** variance)
-    - parameterScaleLaplace: location; scale
+  in parameter scale (e.g. `log10` scale if a parameter has ``parameterScale``
+  `log10`).
+
+  Optionally, a *truncated* distribution can be specified for any distribution
+  except the uniform distributions by providing a lower and upper bound after
+  the prior parameters. For example, a truncated normal distribution would be
```
dweindl, line 843: After dropping `initializationPrior*` (#612), and if we say that the bounds truncate the prior, is there still a need for truncated prior distributions? Not really, right?

sebapersson: Fair point. But then I will make it clear in the documentation that the bounds are truncation points for priors

sebapersson:
Following the reorganization of the parameter table, I have now updated the prior specification. 

Regarding truncated priors, this PR proposes that the parameter bounds define the truncation points.

doc/v2/documentation_data_format.rst
```diff
@@ -792,32 +794,145 @@ Detailed field description
...
-  Prior types used for the :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
+  Prior types used for the 
+  :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
   It is valid to have no priors. However, if priors are specified for a subset
   of parameters, this defaults to the ``uniform`` prior (with ``priorParameters``
   set to ``lowerBound;upperBound``) for the other parameters.
 
-  Possible prior types are:
+  Prior distributions are truncated by the ``lowerBound`` and ``upperBound`` if 
+  the prior's domain is outside the parameter bounds. A non-truncated prior can
```
dweindl, line 804:
```suggestion
  the prior's domain exceeds the parameter bounds. A non-truncated prior can 
```

doc/v2/documentation_data_format.rst
```diff
@@ -792,32 +794,145 @@ Detailed field description
...
+    - *normal*: Normal prior
+    - *rayleigh*, Rayleigh prior
+    - *uniform*: Flat (uniform) prior
 
 - ``priorParameters`` [STRING, OPTIONAL]
 
-  Prior parameters used for the :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
+  Prior parameters used for the 
+  :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
   ``priorParameters`` is required if ``prior`` is non-empty.
```
dweindl, line 827: For `uniform` we could allow empty parameters instead of repeating the bounds?

doc/v2/documentation_data_format.rst
```diff
@@ -792,32 +794,145 @@ Detailed field description
...
 
-  So far, only numeric values will be supported, no parameter names.
+  So far, only numeric values are supported, no parameter names.
```
dweindl, line 829:
```suggestion
  Only numeric values are supported, no parameter names.
```

doc/v2/documentation_data_format.rst
```diff
@@ -153,6 +153,8 @@ PEtab 2.0.0 is a major update of the PEtab format. The main changes are:
   the PEtab format.
 * The admissible values for ``estimate`` in the :ref:`v2_parameters_table`
   are now ``true`` and ``false`` instead of ``1`` and ``0``.
+* Support for new ``prior`` distributions,  as well as support for 
+  truncated distributions in the :ref:`v2_parameters_table`.
```
dweindl, line 157:
```suggestion
* Support for new ``prior`` distributions in the :ref:`v2_parameters_table`, and clarification that bounds truncate the prior distributions.
```

doc/v2/documentation_data_format.rst
```diff
@@ -792,32 +794,145 @@ Detailed field description
...
 
-  Prior types used for the :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
+  Prior types used for the 
+  :ref:`MAP objective function and for Bayesian inference <v2_objective_function>`.
   It is valid to have no priors. However, if priors are specified for a subset
   of parameters, this defaults to the ``uniform`` prior (with ``priorParameters``
   set to ``lowerBound;upperBound``) for the other parameters.
 
-  Possible prior types are:
+  Prior distributions are truncated by the ``lowerBound`` and ``upperBound`` if
```
dweindl, line 803: Do we need to clarify truncation here or is it unambiguous?

dweindl (commented):
Thanks, looks good to me.

Renaming `prior` to `priorDistribution` (https://github.com/PEtab-dev/PEtab/discussions/618#discussioncomment-12628162) could be included here, but can also be done separately.

doc/v2/documentation_data_format.rst
sebapersson: I think it is good to clarify here, so that users easily can identify what to expect when doing, for example, Bayesian inference (and what they need to do if they want unbounded priors)

doc/v2/documentation_data_format.rst
```diff
@@ -148,11 +148,14 @@ PEtab 2.0.0 is a major update of the PEtab format. The main changes are:
   priors are outside the definition of the parameter estimation problem
   and were a source of confusion.
 * ``objectivePriorType`` and ``objectivePriorParameters`` in the
-  :ref:`v2_parameters_table` are renamed to ``prior`` and
+  :ref:`v2_parameters_table` are renamed to ``priorDistribution`` and
   ``priorParameters``, respectively. This change was made to simplify
   the PEtab format.
 * The admissible values for ``estimate`` in the :ref:`v2_parameters_table`
   are now ``true`` and ``false`` instead of ``1`` and ``0``.
+* Support for new ``priorDistribution`` distributions in the
```
dweindl, line 156:
```suggestion
* Support for new parameter prior distributions in the 
```

Maybe clearer, since ``priorDistribution`` didn't exist before.

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
+- *cauchy*:
+
+  - parameters: (location, scale) = :math:`(\mu, \sigma)`
+
+  .. math::
+     \pi(x|\mu, \sigma) = \frac{1}{\pi \sigma \left( 1 + \left(\frac{x - \mu}{\sigma} \right)^2\right)} \quad x \in (-\infty, \infty)
+
+- *chisquare*:
+
+  - parameter: degree of freedom = :math:`\nu`
```
dweindl, line 862:
```suggestion
  - parameter: degrees of freedom = :math:`\nu`
```

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
+  - *uniform*: lower bound; upper bound
+
+Prior distributions
+~~~~~~~~~~~~~~~~~~~
+
+For ``priorDistribution`` several distributions are supported (see above). Denote the parameter
+value by :math:`x` and the Gamma function by :math:`\Gamma`; then, the
+following parameterizations are used:
+
+- *cauchy*:
```
dweindl, line 853: (no) commas separating the domain are inconsistent

dweindl (approved):
:+1:

doc/v2/documentation_data_format.rst
sebapersson: Thanks for noticing, will fix!

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
+    - *cauchy*: Cauchy prior
+    - *chisquare*: Chisquare prior
+    - *exponential*: Exponential prior
+    - *gamma*: Gamma prior
     - *laplace*: Laplace prior
-    - *logNormal*: exponentiated Gaussian prior
-    - *logLaplace*: exponentiated Laplace prior
+    - *log10Normal*: Log10-normal prior
+    - *logLaplace*: Log-Laplace prior
+    - *logNormal*: Log-normal prior
```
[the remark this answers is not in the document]
dweindl: :+1: let's use `log-*`

doc/v2/documentation_data_format.rst
sebapersson: Ahh, you mean also add a link for extra clarification? (I think that would be nice)

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
   Parameters for the different prior types are:
 
-    - uniform: lower bound; upper bound
-    - normal: mean; standard deviation (**not** variance)
-    - laplace: location; scale
-    - logNormal: parameters of corresp. normal distribution (see: normal)
-    - logLaplace: parameters of corresp. Laplace distribution (see: laplace)
+  - *cauchy*: location; scale
+  - *chisquare*: degrees of freedom
+  - *exponential*: rate
```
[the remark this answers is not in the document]
sebapersson: I like this suggestion, as it would remove repeating lists. I think I do not want to include the log-variant columns, as I want to explicitly provide the formula for all distributions (this is helpful for tool developers, in cases they cannot rely on a Distribution package containing a distribution, e.g. LogLaplace is available in Julia yet)

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
+  - *laplace*: location; scale
+  - *log10Normal*: parameters of corresponding normal distribution (see normal)
+  - *logLaplace*: parameters of corresponding Laplace distribution (see laplace)
+  - *logNormal*: parameters of corresponding normal distribution (see normal)
+  - *logUniform*: parameters of corresponding uniform distribution (see uniform)
+  - *normal*: mean; standard deviation (**not** variance)
+  - *rayleigh*: scale
+  - *uniform*: lower bound; upper bound
+
+Prior distributions
```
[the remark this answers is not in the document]
sebapersson: Yes, as commented above I think I will add the table instead

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dilpath: Yes, probably easier to read if the log- variants are separate rows...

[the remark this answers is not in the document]
dweindl: @dilpath added `doc/src/update_tables.py` a while ago to convert tsv tables to rst. This might be handy here. There are some examples in this file already (`START TABLE` ...).

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,147 @@ Detailed field description
...
+  - *exponential*: rate
+  - *gamma*: shape; scale
+  - *laplace*: location; scale
+  - *log10Normal*: parameters of corresponding normal distribution (see normal)
+  - *logLaplace*: parameters of corresponding Laplace distribution (see laplace)
+  - *logNormal*: parameters of corresponding normal distribution (see normal)
+  - *logUniform*: parameters of corresponding uniform distribution (see uniform)
+  - *normal*: mean; standard deviation (**not** variance)
+  - *rayleigh*: scale
+  - *uniform*: lower bound; upper bound
```
FFroehlich, line 844: do we really need extra lower/upper bounds? shouldn't that be implied by parameter bounds?

FFroehlich (approved):

doc/v2/documentation_data_format.rst
sebapersson: This is a good, but also tricky point. You are correct the lower- and upper parameter bounds do imply the bounds. However, I think for consistency that we should for all priors enforce that parameters are specified. I am open for discussing this though.

sebapersson:
Following review suggestions I added the priors in a table instead using the `list-table` syntax, and I think this makes it easier to digest the docs. If everyone agree, I think we can now merge this PR.

doc/v2/documentation_data_format.rst
```diff
@@ -790,34 +793,119 @@ Detailed field description
...
+  * - gamma
+    - shape (:math:`\alpha`); scale (:math:`\theta`)
+    - .. math::
+        \pi(x|\alpha, \theta) = \frac{x^{\alpha - 1}e^{-x/\theta}}{\Gamma(\alpha)\theta^{\alpha}}
+    - :math:`(-\infty, \infty)`
+
+  * - laplace
+    - location (:math:`\mu`); scale (:math:`\sigma`)
+    - .. math::
+        \pi(x|\mu, \sigma) = \frac{1}{2\sigma}\exp\left(- \frac{|x - \mu |}{\sigma}\right) \quad x \in (-\infty, \infty)
```
dilpath, line 864:
```suggestion
        \pi(x|\mu, \sigma) = \frac{1}{2\sigma}\exp\left(- \frac{|x - \mu |}{\sigma}\right)
```

sebapersson: Thanks, missed this :)

doc/v2/documentation_data_format.rst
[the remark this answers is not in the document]
dweindl: > Usually just log-normal or lognormal (i.e., one word, not two words rewritten in camel case).

I changed it to `log-normal` etc..

sebapersson:
This can be merged now right?

PR #601, opened 2025-03-20
Move tool table to RTD

dweindl:
Move the tool support table from the repository README to a separate page of the ReadTheDocs documentation to make the README less crowded. Convert to more maintainable format (auto-generate RST table from TSV).

:eyes: https://petab--601.org.readthedocs.build/en/601/software_support.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,71 +49,9 @@ A wide range of PEtab examples can be found in the systems biology parameter est
 
 ## PEtab support in systems biology tools
 
-Where PEtab is supported (in alphabetical order):
+For a list of tools supporting PEtab, see the
+[PEtab website](https://petab.readthedocs.io/en/latest/software_support.html).
 
-
-  - [AMICI](https://github.com/ICB-DCM/AMICI/)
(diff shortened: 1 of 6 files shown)


doc/v2/documentation_data_format.rst
```diff
@@ -1417,7 +1417,7 @@ The following functions are supported:
      - float
    * - ``sign(x)``
      - | sign of ``x``
-       | ``piecewise(1, x > 0, -1, x < 0, 0)``
+       | ``piecewise(1, x>=0, -1)``
```
dilpath, line 1420: Wikipedia says otherwise, and I don't see a sign function in SBML. What is this based on now?

https://en.wikipedia.org/wiki/Sign_function#Definition

dilpath (approved):
:+1:

doc/v2/documentation_data_format.rst
dweindl: You are right. No intended change, I just regenerated the RST tables. It seems this was fixed in the wrong place last time...

PR #614, opened 2025-03-25
v2: Change parameter.estimate from 0/1 to true/false

dweindl:
Meanwhile, boolean literals are supported in math expressions, so we should also use them for the boolean `estimate` column (see also https://github.com/PEtab-dev/PEtab/discussions/610).

### DIFF
--- a/doc/v2/documentation_data_format.rst
+++ b/doc/v2/documentation_data_format.rst
@@ -143,6 +143,8 @@ PEtab 2.0.0 is a major update of the PEtab format. The main changes are:
   parameter priors was not well-defined.
   To obtain the same effect, the model parameters can be transformed in the
   model file.
+* The admissible values for ``estimate`` in the :ref:`v2_parameters_table`
+  are now ``true`` and ``false`` instead of ``1`` and ``0``.
 
 .. _v2_model:
 .. _v2_model_entities:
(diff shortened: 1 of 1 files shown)


doc/v2/documentation_data_format.rst
```diff
@@ -761,23 +763,24 @@ Detailed field description
...
 
 - ``nominalValue`` [NUMERIC]
 
   Some parameter value to be used if
   the parameter is not subject to estimation (see ``estimate`` below).
-  Optional, unless ``estimate==0``.
+  Optional, unless ``estimate==false``.
 
-- ``estimate`` [BOOL 0|1]
+- ``estimate`` ['true' or 'false']
```
dilpath, line 779: true|false in the table above, but fine for me

dilpath (approved):

doc/v2/documentation_data_format.rst
dweindl, line 779:
```suggestion
- ``estimate`` [``true`` or ``false``]
```

dweindl: Fixed.

FFroehlich (approved):

sebapersson (approved):

PR #616, opened 2025-03-25
v2: Allow applying multiple conditions simultaneously

dweindl:
Allow applying multiple simulation conditions with disjoint changes simultaneously. This will simplify handling factorial designs (see https://github.com/PEtab-dev/PEtab/discussions/615).


doc/v2/documentation_data_format.rst
```diff
@@ -387,8 +388,11 @@ The experiments table has three mandatory columns ``experimentId``,
   Any event triggers defined in the model must also be checked during this
   pre-simulation.
 
-  Multiple conditions cannot be applied simultaneously within a single
-  experiment; each ``time`` value must be unique for each ``experimentId``.
+  Multiple conditions may be applied simultaneously by specifying the same
+  ``experimentId`` and ``time`` for multiple conditions. The changes associated
+  with these conditions must be disjoint, i.e., no two conditions applied at
+  the same time may modify the same entity. The union of these changes is
+  applied to the model as if they were specified as a single condition.
```
dilpath, line 395:
```suggestion
  Multiple conditions may be applied simultaneously by specifying the same
  ``experimentId`` and ``time`` for multiple conditions. The changes associated
  with these conditions must be disjoint, i.e., no two conditions applied at
  the same time may modify the same ``targetId``. The union of these changes is
  applied to the model as if they were specified as a single condition.
```

dilpath (approved):
:+1:

sebapersson (approved):

doc/v2/documentation_data_format.rst
dweindl: Just "target"? We are not modifying the targetId as such.

dilpath: True, reworded slightly now, but fine to also just use target.

Issue #609, opened 2025-03-24
Clarify PEtab problems with multiple models


dweindl:
So far, the [spec says](https://petab.readthedocs.io/en/latest/v1/documentation_data_format.html#parameter-estimation-problems-combining-multiple-models):

> Parameter estimation problems can comprise multiple models. For now, PEtab allows to specify multiple SBML models with corresponding condition and measurement tables, and one joint parameter table. This means that the parameter namespace is global. Therefore, parameters with the same ID in different models will be considered identical.

I don't think this is sufficient to implement support for that.

Clarify, for example:

* Is there is a global namespace for conditions/experiments/... across that `problems` list in the yaml file?
* Can a measurement file from `problems[0]` refer to observables/experiments from `problems[1]`, and so on. If so, how to specify to which model an observable / a measurement belongs?
* ...

dweindl:
#622

PR #622, opened 2025-04-22
v2: Clarification MultiModel format

FFroehlich:
See https://github.com/PEtab-dev/PEtab/issues/609

One could also think about adding the option to specify models for individual conditions, but that definitely sounds like extension territory.


doc/v2/documentation_data_format.rst
```diff
@@ -185,10 +186,10 @@ PEtab distinguishes between three types of entities:
...
 ----------------
 
-The optional conditions table defines discrete changes to the model. These (sets of)
-changes typically represent interventions, perturbations, or changes in the
-environment of the system of interest. These modifications are referred to as
-(experimental) *conditions*.
+The optional conditions table defines discrete changes to the simulated model. 
+These (sets of) changes typically represent interventions, perturbations, or 
+changes in the environment of the system of interest. These modifications are 
+referred to as experimental) *conditions*.
```
dilpath, line 192:
```suggestion
referred to as (experimental) *conditions*.
```

doc/v2/documentation_data_format.rst
```diff
@@ -1033,12 +1055,14 @@ easy validation:
...
-will be considered identical.
-
+Parameter estimation problems can involve multiple models, with a specific model
+assigned to each individual data point. However, all tables in a PEtab problem apply
+uniformly to all models. As a result:
+- Multiple models may need to be simulated for a single experiment.
+- Each model may be simulated for a distinct subset of experiments.
+- The number of conditions that must be simulated for an experiment may vary depending 
+  on the model used.
+- Each parameter in the parameter table hase the same value across all models.
```
dilpath, line 1065:
```suggestion
- Each parameter in the parameter table has the same value across all models.
```

dilpath (approved):
- There is a small inconsistency, since the "PEtab" way would be to have a new `models` table, I think.

- Instead of a `modelId` column, the user could instead associate one measurement file with one model file in the YAML directly.

But the suggestion is fine for me, especially since I have no use cases currently.

matthiaskoenig (commented):
Not sure if adding the modelId to the measurementTable is the best solution.
I was thinking about adding the `modelId` column to the `conditionTable`, `parameterTable` and `observableTable`. I.e. wherever an id is referenced, it should be possible to provide an optional `modelId` to specify which model the id should be from. This would allow you to validate the table against the model. Otherwise you have to go back from the measurementTable to the other table to figure out which model was meant for which variable. This makes it very difficult to work with multiple models.

I could imagine having a problem with multiple models where parameters from each model are used, as well as observables from each model, and in the time series conditions are applied to each model. One wants to see from which model each information is coming in the respective table.

dweindl:
> * There is a small inconsistency, since the "PEtab" way would be to have a new `models` table, I think.

I'd see the yaml file as kind of the "model table" here:

```yaml
# [...]
  model_files:
    model1:
      language: sbml
      location: model1.xml
    model2:
      language: sbml
      location: model2.xml
# [...]
```

dweindl:
> * Instead of a `modelId` column, the user could instead associate one measurement file with one model file in the YAML directly.

That was the original idea behind the `problems` list in the yaml file:

```yaml
format_version: 2.0.0
parameter_file: all_parameters.tsv
problems:
- condition_files: [conditions_for_model1.tsv]
  experiment_files: [experiments_for_model1.tsv]
  measurement_files: [measurements_for_model1.tsv]
  model_files:
    model1:
      language: sbml
      location: model1.xml
  observable_files: [observables_for_model1.tsv]
- condition_files: [conditions_for_model2.tsv]
  experiment_files: [experiments_for_model2.tsv]
  measurement_files: [measurements_for_model2.tsv]
  model_files:
    model2:
      language: sbml
      location: model2.xml
  observable_files: [observables_for_model2.tsv]
  ```

However, I am not sure what is preferable.

File: doc/v2/documentation_data_format.rst [hunk 3]
```diff
@@ -593,12 +612,15 @@ Detailed field description
...
-  May contain any symbol defined in the SBML model (including model time ``time``)
-  or parameter table. In the simplest case just an SBML species ID
-  or an ``AssignmentRule`` target. Additionally, any observable ID
+  May contain any symbol defined any of the employed models (including model 
+  time ``time``) or parameter table. In the simplest case, for SBML models, 
+  just a species ID or an ``AssignmentRule`` target. Additionally, any observable ID
   introduced in the observable table may be referenced, but circular definitions
   must be avoided.
 
+  In problems with multiple models, symbols not defined in the currently simulated
```
sebapersson, line 621: Should this not rather throw an error?
If I for example there is in the observable formula a specie not present in the model, it feels like an error has been made by the user.

doc/v2/documentation_data_format.rst
```diff
@@ -758,7 +780,7 @@ Detailed field description
 - ``parameterId`` [PETAB_ID, REQUIRED]
 
   The ``parameterId`` of the parameter described in this row. This has to match
-  the ID of a parameter specified in the SBML model, a parameter introduced
+  the ID of a parameter specified any model, a parameter introduced
```
sebapersson, line 783:
```suggestion
  the ID of a parameter specified at least one model, a parameter introduced
```

doc/v2/documentation_data_format.rst
```diff
@@ -990,8 +1012,8 @@ Detailed field description
 
 - ``modelEntityId`` [STRING or empty, REQUIRED]
 
-  A globally unique identifier defined in the model, or empty if the entity is
-  not present in the model. This does not have to be a valid PEtab identifier.
+  A globally unique identifier defined in any model, or empty if the entity is
+  not present any model. This does not have to be a valid PEtab identifier.
```
sebapersson, line 1016:
```suggestion
  not present in any model. This does not have to be a valid PEtab identifier.
```

sebapersson:
> I was thinking about adding the `modelId` column to the `conditionTable`, `parameterTable` and `observableTable`. I.e. wherever an id is referenced, it should be possible to provide an optional `modelId` to specify which model the id should be from. This would allow you to validate the table against the model. Otherwise you have to go back from the measurementTable to the other table to figure out which model was meant for which variable. This makes it very difficult to work with multiple models.

If `modelId` is added to the `parameterTable`, would it not then be quite involved to have the same parameter in multiple models (e.g. one would have to add multiple model-ids?). I think having the modelId only in the measurements table is doable, however, the drawback as pointed out here is that it will be quite a bit of work for the importer to figure out which parameters and species appear in each model.

matthiaskoenig:
> That was the original idea behind the problems list in the yaml file.

As I understand it one will have a single optimization problem which uses two models. I.e. in the end there is a single cost function which has contributions of the different models. Having a list of problems implies these are separate optimization problems with multiple cost functions which could be optimized independently. 

> If modelId is added to the parameterTable, would it not then be quite involved to have the same parameter in multiple models (e.g. one would have to add multiple model-ids?).

That is a good point. I think we have to write down clearly the use cases we want to implement with the multiple models and then figure out the best way on how to encode this. 
Personally, I am always more in favour of having a more fine grained control (i.e. by adding model ids to the different tables) which allows in the long run much more flexibility and implement use cases nobody thought about. By adding the modelId on the measurements table we allow only to distribute measurements to different models, but nothing else. By having modelIds on the separate tables a lot of different things could be done in the future without changing the format much. E.g. adding multiple modelIds to indicate that parameters or conditions should be shared between models.

My use cases are:
1. **Smaller models by distribution of problem in submodels (shared parameters).** I have multiple models which map to different datasets, and there are some parameters which occur in the different models (shared parameters). I want to optimize some of the parameter which occur in all models as a single parameter (e.g. hepatic blood flow, ...). Allowing to split the problem makes the ODE integration of the submodels and code generation much faster. Code generation via LLVM and other C++ code optimization does not scale linearly, i.e. code generation scales > 0(n). These becomes the bottleneck for large models (we have to wait >12 hours for some code generation with roadrunner with simulation taking a few seconds if we have > 100 000 state variables).
2. **Shared experiments between models.** I have different models and want to run the same experiments over it. No shared parameters, but shared experiments. Could be encoded as separate optimization problems. Duplicating all the information is error prone. Similar as 3. 
3. **Encoding multiple optimization problems in a single file**. I want to change parameters/settings of the optimization problem/model and run all the variants. Something in the direction of a sensitivity analyses to settings (hyperparameter scans). E.g. test out different rate laws, different prioris, ... I could generate a lot of different PETab problems, but it would be much nicer to encode these in a single PETab by providing the different models. This would require different cost functions for the different models/settings and goes in the direction of model selection. This does not require multiple models in a single problem, but multiple problems in a single file. If one has larger models a ton of unnecessary data is generated with the model files being the largest assets for the optimization problem.

For me the main reason is to 1. be able to split large models in submodels for part of the cost function calculation. This is mainly an optimization aspect and could perhaps also be done on a tool bases if there is some metadata. 2. Encode multiple problems in a single PETab problem (this is mainly convenience and reuse of information). Basically a level on top of a single optimization problem.

matthiaskoenig:
Optimizing a single parameter from multiple models could require a mapping of the parameters, because the parameters do not have to be named identically in the different models (especially if existing models from different sources are used). The more I think about the multiple model problems the more complicated it gets. We really have to think this through.

FFroehlich:
> Optimizing a single parameter from multiple models could require a mapping of the parameters, because the parameters do not have to be named identically in the different models (especially if existing models from different sources are used). The more I think about the multiple model problems the more complicated it gets. We really have to think this through.

This would already be covered by the implementation above with the mapping performed via the condition table. Things get problematic if the same parameter IDs are reused across models but with different meanings, but at the end of the day that's just asking for trouble.

doc/v2/documentation_data_format.rst
FFroehlich: Yes, I considered that alternative. This specification allows reuse of observables across models, the user can always introduce model specific observables. We could provide different options for linting that are more or less strict.

doc/v2/_static/petab_schema_v2.yaml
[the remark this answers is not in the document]
FFroehlich: briefly, removed the problems list/array and moved all properties to the highest level.

doc/v2/documentation_data_format.rst
```diff
@@ -785,7 +786,7 @@ Detailed field description
 - ``parameterId`` [PETAB_ID, REQUIRED]
 
   The ``parameterId`` of the parameter described in this row. This has to match
-  the ID of a parameter specified any model, a parameter introduced
+  the ID of a parameter specified at least one model, a parameter introduced
```
dilpath, line 789:
```suggestion
  the ID of a parameter specified in at least one model, a parameter introduced
```

doc/v2/documentation_data_format.rst
```diff
@@ -1145,14 +1146,51 @@ easy validation:
...
+- A single experiment may require simulations using multiple models.
+- Each model may be associated with a distinct subset of experiments.
+- The number of conditions to be simulated for a model-specific instance
+of an experiment may vary across models.
+- Each parameter defined in the :ref:`v2_parameters_table` has a shared value across all 
+models.
+
+Validation Rules
+^^^^^^^^^^^^^^^^
+For any given model Only those experiments and observables that appear in the
```
dilpath, line 1186:
```suggestion
For any given model, only those experiments and observables that appear in the 
```

dilpath (approved):
Fine for me, no strong opinion for or against

doc/v2/documentation_data_format.rst
```diff
@@ -304,6 +304,9 @@ phases, following the logic of the event assignments in a single SBML event.
      For further details, refer to SBML semantic test suite case `01779
      <https://github.com/sbmlteam/sbml-test-suite/blob/7ab011efe471877987b5d6dcba8cc19a6ff71254/cases/semantic/01779/01779-model.m>`_.
 
+   * For problems involving multiple models, any assignment to targets not
+     present in the currently simulated model will be ignored.
```
dweindl, line 308: I am not sure that will solve more problems than it creates, but fine if there are good use cases.

dweindl (commented):
While I am happy to get rid of the `problems` list in the yaml file for the single-model case, I thought the editor meeting discussion ended with going back the old style of having different sets of condition, observables, ... files for each model with only the parameter table being shared. (Which would not preclude referencing the same observable/condition/experiment file from multiple sub-problems). 

I can live with either and would follow those who already have a clear application in mind.

doc/v2/documentation_data_format.rst
FFroehlich: sorry, this should have been deleted.

FFroehlich:
> While I am happy to get rid of the `problems` list in the yaml file for the single-model case, I thought the editor meeting discussion ended with going back the old style of having different sets of condition, observables, ... files for each model with only the parameter table being shared. (Which would not preclude referencing the same observable/condition/experiment file from multiple sub-problems).
> 
> I can live with either and would follow those who already have a clear application in mind.

That wasn't my takeaway. I thought we mainly discussed about how forgiving conditions/observables are in terms of validation.

Thinking about it, splitting up the problem at the file level is actually simpler to implement and more verbose for the user, so that definitely is a viable alternative that is equivalent in terms of what can be expressed. But yes, would be great if we don't have a problems list for the single-model case.

We could drop the model id column from the measurements table and introducing an additional sub-problems table with model id as columns and files as rows with True/False values to indicate which file applies to which model? Default would be to apply tables to all models, minimising the number of rows that have to be added.

doc/v2/documentation_data_format.rst
```diff
@@ -975,11 +984,53 @@ easy validation:
...
+
+- Reuse of experiments across models.
+- Fine-grained model-to-data assignment.
+
+With the exception of the :ref:`v2_measurements_table`, all other PEtab tables apply
+uniformly to all models.
+
+This design has several implications:
+
+- A single experiment may require simulations using multiple models.
```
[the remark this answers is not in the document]
FFroehlich: > I find this potentially confusing. The message is that a given experiment may have to be simulated (completely) once with model A and once with model B in order to evaluate the objective function, but not that period 1 of a given experiment is simulated with model A, and period 2 is simulated with model B, right?

correct.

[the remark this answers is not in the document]
dweindl: I'd specifically add that this is decided based on the measurements.

"A single experiment may need to be simulated with different models for different measurements. However, a single simulation of a given experiment is always performed using one single model."?

dweindl:
@matthiaskoenig @sebapersson : Does this work for you?

doc/v2/documentation_data_format.rst
```diff
@@ -531,6 +539,14 @@ Detailed field description
...
   ``observableParameters`` in the previous point.
 
+- ``modelId`` [PETAB_ID, OPTIONAL, REFERENCES(yaml.models.model_id)]
+
+  The modelId specifies which model to simulate for each data point. modelIds
+  are defined by the keys of the models dict in the PEtab problem YAML file.
+  This column is required when multiple models are defined in the PEtab problem.
+  For problems with a single model, this column is ignored, and the same model
+  is used for all simulations.
+
```
sebapersson, line 549:
```suggestion
  Which model to simulate for each data point. modelIds are defined by the keys of 
  the models dict in the PEtab problem YAML file. This column is required when 
  multiple models are defined in the PEtab problem, otherwise it is ignored.
```

sebapersson (approved):
Looks good for me. Minor comments.

For implementing support for this feature, it would be nice to have test-cases covering this.

doc/v2/_static/petab_schema_v2.yaml
```diff
@@ -22,96 +22,67 @@ properties:
...
+        type: object
+        properties:
+          location:
             type: string
-            description: PEtab condition file name or URL.
+            description: Model file name or URL
+          language:
+            type: string
+            description: |
+              Model language, e.g., 'sbml', 'cellml', 'bngl', 'pysb'
```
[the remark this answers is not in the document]
FFroehlich: well sbml is also formatted in xml 🤷‍♂️

doc/v2/documentation_data_format.rst
```diff
@@ -531,6 +539,14 @@ Detailed field description
   Numeric values or parameter names are allowed. Same rules apply as for
   ``observableParameters`` in the previous point.
 
+- ``modelId`` [PETAB_ID, OPTIONAL, REFERENCES(yaml.models.model_id)]
+
+  The modelId specifies which model to simulate for each data point. modelIds
+  are defined by the keys of the models dict in the PEtab problem YAML file.
+  This column is required when multiple models are defined in the PEtab problem.
+  For problems with a single model, this column is ignored, and the same model
+  is used for all simulations.
```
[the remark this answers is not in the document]
FFroehlich: Yes.

doc/v2/_static/petab_schema_v2.yaml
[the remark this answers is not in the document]
sebapersson: > PySB models directly, I'd still prefer they are able to use PEtab, so that at least the rest is more standardized and portable.

Feels like this should then rather be up to implementations? Otherwise it would also be fair to mention frameworks as Catalyst.jl and ModelingToolkit.jl here (which PEtab.jl supports). So how about we just remove pysb here?

doc/v2/documentation_data_format.rst
```diff
@@ -1013,15 +1029,61 @@ easy validation:
...
+
+- A single experiment may need to be simulated with different models for
+  different measurements.
+  However, a single simulation of a given experiment is always performed
+  using one single model
+- Each model may be associated with a distinct subset of experiments.
+- The number of conditions to be simulated for a model-specific instance
+  of an experiment may vary across models.
+- Each parameter defined in the :ref:`v2_parameters_table` has a shared value
+  across all models.
```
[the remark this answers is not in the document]
dilpath: Thanks, B sounds fine to me too. This is like how a model component can take different values across conditions. In v1, a simulation is a unique combination of (preequilibration condition, simulation condition). In v2, it will be a unique combination of (experiment, model). Parameter/state variable values may differ between simulations in v2, due to differences in the experiment or model.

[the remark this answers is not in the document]
dweindl: Agreed, B sounds most reasonable.

C would be a bit weird. It would be illegal to have a constant SBML parameter with different values in different models, but it would be fine to have a constant SBML species with different values in different models.

dilpath (approved):
:+1:

doc/v2/_static/petab_schema_v2.yaml
```diff
@@ -22,96 +22,67 @@ properties:
...
-
-          items:
-            type: string
-            description: PEtab measurement file name or URL.
-
-        condition_files:
-          type: array
-          description: List of PEtab condition files.
+  model_files:
+    type: object
```
matthiaskoenig, line 27: why is this `object` but `array` for the other files? Should this not be `array`

doc/v2/documentation_data_format.rst
```diff
@@ -205,10 +213,10 @@ PEtab distinguishes between three types of entities:
 Conditions table
 ----------------
 
-The optional conditions table defines discrete changes to the model. These (sets of)
-changes typically represent interventions, perturbations, or changes in the
-environment of the system of interest. These modifications are referred to as
-(experimental) *conditions*.
+The optional conditions table defines discrete changes to the simulated model.
```
matthiaskoenig, line 216: `model` -> `model(s)`

matthiaskoenig (changes requested):
Some minor comments to be addressed.

doc/v2/_static/petab_schema_v2.yaml
```diff
@@ -22,96 +22,67 @@ properties:
...
+    description: One or multiple models
 
-          items:
+    # the model ID
+    patternProperties:
+      "^[a-zA-Z_]\\w*$":
```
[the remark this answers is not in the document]
dweindl: It's correct. This is just a regular string. It is interpreted as a regex only elsewhere.

doc/v2/_static/petab_schema_v2.yaml
dweindl: So far it's:
```yaml
model_files:
  some_id:
    location: ...
    language: ...
```
This could be changed to 

```yaml
model_files:
  - id: some_id
    location: ...
    language: ...
```

I am fine with either, but it's out of scope of this PR.

doc/v2/_static/petab_schema_v2.yaml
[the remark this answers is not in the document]
dweindl: I am fine with about anything here, as long as `sbml` stays in there. However, I am very much in favor of resolving that somewhere else. This has been in place since #538 and should not block this pull request.

doc/v2/documentation_data_format.rst
```diff
@@ -1013,15 +1029,66 @@ easy validation:
...
+
+This design has several implications:
+
+- A single experiment may need to be simulated with different models for
+  different measurements. However, a single simulation of a given experiment
+  is always performed using one single model.
+- Each model may be associated with a distinct subset of experiments.
+- The number of conditions to be simulated for a model-specific instance
+  of an experiment may vary across models.
+- Each parameter defined in the :ref:`v2_parameters_table` has a shared value
```
sebapersson, line 1077: Worth to mention that we can achieve model specific parameters via using the condition table (we can map parameter k_m1, k_m2 to k in the condition table, where k then can set different values in different models)

sebapersson (approved):
I think it looks good, left a minor comment.

Probably outside the scope of this PR, but it can be worthwhile to somewhere note down how to implement this in a tool. How I see it, the easiest approach is that each model can be made to its own PEtab problem, then there is only a need for a master PEtab problem to delegate correct parameters to each sub-problem. This would, for example, in an implementation allow specific gradient methods to be set for each model type, which would benefit performance.  Creating each "sub" PEtab problem could maybe be something the PEtab library supports?

FFroehlich:
> I think it looks good, left a minor comment.
> 
> Probably outside the scope of this PR, but it can be worthwhile to somewhere note down how to implement this in a tool. How I see it, the easiest approach is that each model can be made to its own PEtab problem, then there is only a need for a master PEtab problem to delegate correct parameters to each sub-problem. This would, for example, in an implementation allow specific gradient methods to be set for each model type, which would benefit performance. Creating each "sub" PEtab problem could maybe be something the PEtab library supports?

How about creating a separate document for that? Would prefer to keep recommendations for developers out of the spec as our primary target audience should be end-users.

doc/v2/documentation_data_format.rst
sebapersson: I am not 100% sure users not experienced with PEtab will make the connection they can use the condition table to assign model specific values for a SBML model parameters, e.g. k1, via the condition table. This is also just a suggestion, which if you do not think is needed, can be ignored.

sebapersson: I think that is an excellent point that would help users create more tricky PEtab setups (we should probably open a separate issue for this)!

dweindl:
ping @matthiaskoenig @fbergmann

Issue #626, opened 2025-05-07
v2: add clarification sentence about simulation start time and end times


matthiaskoenig:
We should add a clarification sentence about what are the start time and end time for simulations. This is important for time triggered events and be more explicit.

- simulations should start from the first time given in the experiments table or time of the first measurement for that experiment in the measurements table (i.e. what is earlier)
- and run until the last time given in the experiments table or the time of the last measurement for that experiment in the measurements table (i.e. what is later)
- If no experiment is specified, the model simulation is assumed to start at time 0, unless the model specifies a different initial time and runs until the last measurement for that experiment.

Personally I would prefer that we enforce that the first time given in the experiments table must be <= the time of the first measurement and the last time must be >= the time of the last measurement. Than the start and end time of the simulation would be clearly defined.

dweindl:
> We should add a clarification sentence about what are the start time and end time for simulations. This is important for time triggered events and be more explicit.

I think it's rarely needed, but I agree that specifying an end time different from the last measurement could be relevant sometimes. E.g., situations where one needs to track some max value across a certain period extending beyond the last measurement.

>   * simulations should start from the first time given in the experiments table or time of the first measurement for that experiment in the measurements table (i.e. what is earlier)

I would consider a measurement time before the start of the experiment an error, see also below.


>   * If no experiment is specified, the model simulation is assumed to start at time 0, unless the model specifies a different initial time and runs until the last measurement for that experiment.

Start is already specified: 

https://github.com/PEtab-dev/PEtab/blob/4298e15aa9d03e46d7dc6706ebbdf12ddc5a1030/doc/v2/documentation_data_format.rst?plain=1#L486-L488

> Personally I would prefer that we enforce that the first time given in the experiments table must be <= the time of the first measurement 

Agreed. I think this was everybody's assumption, but I can't find it written anywhere.

>   * and run until the last time given in the experiments table or the time of the last measurement for that experiment in the measurements table (i.e. what is later)

> and the last time must be >= the time of the last measurement. Than the start and end time of the simulation would be clearly defined.

For any given experiment, the simulation must be run until (including) the timepoint of the last measurement or the beginning of the last period, whichever comes later (no shorter, no longer). Works for me.

dweindl:
> I think it's rarely needed, but I agree that specifying an end time different from the last measurement could be relevant sometimes.

[...]

PR #630, opened 2025-05-21
v2: Clarify valid measurement timepoints

dweindl:
Related to #626.


dilpath (approved):
Fine for now but I would prefer clarity through some figure, TODO

PR #632, opened 2025-05-21
v2: Clarify simulation file

### LINKED ISSUE #628: specify column name for simulation data for optional simulation.tsv
**Which problem would you like to address? Please describe.**
Specs for v1 and v2 mention that we can have an optional simulation data file "in the same format as the measurement" file, however whenever they are used the measurement column has a different name. That name has not been consistent in the past it was either 'simulatedData` or `simulation`. 

**Describe the solution you would like**
Add a clarification that the format has to be the same as the measurement file, with `measurement` replaced by `simulation`

dweindl:
* Remove the simulation from the list of PEtab files, because it is not part of the parameter estimation problem.
* Add a section under *measurement table* stating the expected column name.

Closes #628.


doc/v2/documentation_data_format.rst
```diff
@@ -545,6 +542,20 @@ Detailed field description
...
 
+.. _v2_simulation_table:
+
+Simulation table
+~~~~~~~~~~~~~~~~
+
+For some applications, it is useful to provide a simulation file that contains
+the simulation results of the model for the same time points as the
+measurements. This is useful for example for comparing the simulation results
+with the measurements or for distributing expected model outputs.
```
dilpath, line 553: Makes sense as is, but this is clearer for me.

```suggestion
measurements. This is useful, for example, for comparing the simulation results
with the measurements, or for assessing simulation reproducibility across tools.
```

dilpath (approved):

sebapersson:
> To be more useful, I think also the actual noise value should be part of that table, at least for those cases where noiseFormula depends on the model state. But currently, I think, everything assumes that noiseParameters is a numeric value, or at least only depends on (constant) parameters.

This is a good point that especially can help with testing. If it is included I think it is should be included for all measurement points.

dweindl:
Including the evaluated noise scale would require an additional column (`simulationNoise`?). Just writing that value to `noiseParameters` would be incorrect in all cases in which 'noiseFormula` is not just a single parameter, and not having the original `noiseFormula` would make it impossible to match the rows from the measurement table sometimes.

dweindl:
> Maybe easiest to not include it in the current iteration?

I'd rather remove any mention of the simulation file then until we converged on what it should look like. I am not sure it's helpful if different tools have different expectations of what columns are present...

sebapersson:
I do not even know if most tools support outputting the noise, while due to the test-suite most tools should be able to output simulated values. So feels it should be fine to keep simulation as it is, and in future iterations add noise-column potentially?

PR #633, opened 2025-06-05
v2: Update PEtab files figure

dweindl:
Updated PEtab files figure to match the current v2 draft.

Opinions?

- [x] pdf, png


dilpath (changes requested):
Looks nice!

- [x] text on left side is grey, but not easily noticeable. Make it lighter grey like "Visualization" in the v1 figure? Or group them with a similar background color and title the area "Optional tables"?

I'll send you a copy with some wording suggestions too

dilpath (approved):

m-philipps:
What do you think of reversing the order of the tiles, so the optional files are to the right? 
I also greyed out the arrows pointing to the optional files. If the order does not matter, `Parameters` and `Measurements` could be switched.
petab_files_rev

matthiaskoenig:
+1 I prefer that.

dweindl:
> What do you think of reversing the order of the tiles, so the optional files are to the right?

I guess it depends on how you read the figure. If you read left-to-right, yes, it makes sense. I read it clockwise starting from the PEtab logo, though, and found it more intuitive before.


> I also greyed out the arrows pointing to the optional files. 

Fine with me.

> If the order does not matter, `Parameters` and `Measurements` could be switched.

Fine with me.

Issue #624, opened 2025-05-05
v2: Clarify condition table & initial assignments


dweindl:
A question regarding the current v2 draft and how initial assignments interact with changes in the condition table:

https://github.com/PEtab-dev/PEtab/blob/4298e15aa9d03e46d7dc6706ebbdf12ddc5a1030/doc/v2/documentation_data_format.rst?plain=1#L274-L305

If we have a model with a parameter  `p0 = 1` and another parameter `p` that is the target of an initial assignment `p <- p0`, and the experiment/condition tables specify some change `p0 = 2` at time = 0 -- what is the value of `p` at time=0? 1 or 2?

According to the current draft, I would say the value should be `1`, because the initialAssignment is evaluated before the condition-specific changes are applied.
This would be in line with how events triggering at t=0 would be handled in SBML, i.e. after the evaluation of initialAssignments. However, I am not sure if this is what users would expect. I think, in PEtab v1 this wasn't clearly specified, but implementations and users (and the test suite) assumed that changes through the condition table would be applied before evaluating initialAssignments (and thus, for the example above, the value of `p` would be `2`). 

How should this be handled? Generally, I'd find applying the condition table changes before evaluating initialAssignments more convenient. However, this is problematic if the targetValues themselves depend on targets of initialAssignments.

dilpath:
In v1, if a species was assigned a value in the conditions table, this was assumed to override the `initialAssignmentRule` for that species, I think?

> If a species ID is provided, it is interpreted as the initial condition of that species (as amount if hasOnlySubstanceUnits is set to True for the respective species, as concentration otherwise) and will override the initial condition given in the SBML model or given by a preequilibration condition. If NaN is provided for a condition, the result of the preequilibration (or initial condition from the SBML model, if no preequilibration is defined) is used.

This is at least true for AMICI: https://github.com/AMICI-dev/AMICI/blob/a87058e94831700fa18b8922e5ba33adb22fec91/python/sdist/amici/petab/sbml_import.py#L130-L135

If so, then overriding initialAssignmentRules for parameters would also make sense.

dweindl:
> In v1, if a species was assigned a value in the conditions table, this was assumed to override the `initialAssignmentRule` for that species, I think?

[...]

PR #645, opened 2025-07-24
v2: Clarify model initialization

dweindl:
Clarify that the parameter table is applied before initial assignments are executed.

### DIFF
--- a/doc/v2/documentation_data_format.rst
+++ b/doc/v2/documentation_data_format.rst
@@ -307,6 +307,12 @@ phases, following the logic of the event assignments in a single SBML event.
    * For the initial time period of an experiment, *current* values are
      determined by the initial conditions defined in the model.
 
+     For the all parameters that occur in the parameter table, the
+     respective externally provided values for any estimated parameters,
+     or ``nominalValue`` for any non-estimated parameters are used for model
+     initialization. Any model constructs that would override the initial
+     values of a model entity that occurs in the parameter table are ignored.
+
      Constructs defining initial values of model entities (e.g., SBML's
      *initialAssignments*) are applied only once at the start of the first
      time period and are not re-evaluated in later time periods. The start


matthiaskoenig (changes requested):
I'm not sure if this clarifies things.

In my opinion, InitialAssignments must be applied when the model is loaded, i.e., before any changes to the model are made. This is how I understand SBML and SED-ML handle InitialAssignments.

All InitialAssignments are applied when the model loads.

Many tools eliminate InitialAssignments when loading a model because they apply the changes and then leave the model in an initialized state. Roadrunner does something similar.
To override initial assignments, the new assignment must be encoded in the parameter table.

In summary, we have to clarify this. However, my interpretation is different; InitialAssignments must be applied first.

matthiaskoenig:
To clarify. The changes in the parameter table are basically EventAssignments (and are handled as such in the timecourse definition). EventAssignments are applied after InitialAssignments.

See SBML Specification for simulation at t=0  (https://raw.githubusercontent.com/combine-org/combine-specifications/main/specifications/files/sbml.level-3.version-2.core.release-2.pdf, p29 L10):
* Any InitialAssignment definition continues to take effect, but may now be overridden by any
EventAssignment,

dweindl:
Yes, you are right. That was too hasty. My main problem was that so far, it was not mentioned at all how parameter table parameters are treated. Changing a parameter for the first period via the condition table should have the same effect as changing it through the parameter table. It can be phrased in a way that the parameter table implicitly defines a default condition that is applied to the first period of every experiment.

dweindl:
Hm, I think we'd a problem there if we consider the parameter table parameters part of the first condition to be applied:

Assume we have a species `x1`, and we want to estimate its initial value `x1_init`. In the prevalent petab v1 interpration,  one would have set `x1 = x1_init` in the condition table. 
This is no longer possible because:

> targetValues are first evaluated using the current values of all variables in the respective expressions.

> For the initial time period of an experiment, current values are determined by the initial conditions defined in the model.

This way, `x1 = x1_init` in the condition table would not use the current value of the estimated `x1_init` but its original `x1_init` value from the model (if it exists --- otherwise undefined behavior).

Thus, we need to include an additional initialization step:

1. Model-interal initialization (ignore anything PEtab)
2. **Apply parameter table parameters**
3. Evaluate targetValues for the first period
4. Continue as described in the current spec

dilpath:
I see parameter table parameter values as changing the model definition before any simulation is initialized. This way:
- initial conditions $x_0$ are computed with the parameter table values, if $x_0$ is specified as initialAssignmentRules in SBML. Parameter table parameters are not timecourse parameters, so they are not similar to SBML events.
- in gradient-based optimization, the sensitivity of the initial condition to the estimated parameter $\frac{\partial x_0}{\partial \theta}$ will be computed with the estimated parameter value, as opposed to the model value. If the initial condition is defined in the model by initialAssignmentRules, as opposed to the PEtab condition table, the parameter table values will never be used for $x_0$ or $\frac{\partial x_0}{\partial \theta}$ -- but they should be.

FFroehlich:
> > * Parameter changes won’t affect initial conditions, which I find counterintuitive.
> > * Having parameter changes affect initial conditions would require moving all relevant initial assignments to the condition table, which reduces flexibility.
> 
> I fully agree with that. That's similar to the discussion in #624. However, there the conclusion was that condition table changes should be treated as event assignments, including the changes at the initial timepoint, and that moving all initialization to the condition table was fine.
> 
> In PEtab v1, changing some entity through the condition table had the same effect as changing it through the parameter table. I would greatly prefer keeping it that way. I'd find it very confusing if setting A0=0 in the parameter table affected my initial assignments, but specifying the same in the condition table for all experiments would be ignored in my initial assignments.
> 
> So I'd say parameter table and condition table changes should both be applied before initial assignments, or both after, but not mixed.

Fair point. The above example also applies to setting where you want to map different initial condition parameters, which also wouldn't work with the current spec. Sounds we either need revisit that whole topic again or have a big massive warning sign because this is something user, including me, will get wrong.

dweindl:
> Rereading that whole thread, I would say let's go with the big fat warning signs, i.e. check whether any parameter/condition table assignments appear in the initial assignments.

Although I'm not so much looking forward to go over all test cases again, my conclusion would have been the opposite :see_no_evil:  I can live with either, though.

Since this is quite a central aspect, it would be great to get some additional opinions.

dweindl:
> I see parameter table parameter values as changing the model definition before any simulation is initialized. This way:
> - initial conditions $x_0$ are computed with the parameter table values, if $x_0$ is specified as initialAssignmentRules in SBML. Parameter table parameters are not timecourse parameters, so they are not similar to SBML events.

I'd say this depends on how you define $x_0$. If $x_0 = x(t_0, \theta)$, then this would be the post-event state. If you don't want to consider the condition table for computing $x_0$, this means that there can't be a condition-specific $x_0(\theta)$. I don't think that's what we want.

> - in gradient-based optimization, the sensitivity of the initial condition to the estimated parameter $\frac{\partial x_0}{\partial \theta}$ will be computed with the estimated parameter value, as opposed to the model value. If the initial condition is defined in the model by initialAssignmentRules, as opposed to the PEtab condition table, the parameter table values will never be used for $x_0$ or $\frac{\partial x_0}{\partial \theta}$ -- but they should be.

Hm, in gradient-based optimization, we usually don't really care about $\frac{\partial x_0}{\partial \theta}$, but only about $\frac{\partial \mathcal{L}}{\partial \theta}$, so this wouldn't really be argument for or against either, would it?

dilpath:
> > * Parameter changes won’t affect initial conditions, which I find counterintuitive.
> > * Having parameter changes affect initial conditions would require moving all relevant initial assignments to the condition table, which reduces flexibility.
> 
> I fully agree with that. That's similar to the discussion in #624. However, there the conclusion was that condition table changes should be treated as event assignments, including the changes at the initial timepoint, and that moving all initialization to the condition table was fine.

I voted for condition changes before the initial condition there :grinning: There was a lot of back and forth but again here I would say both condition changes and parameter table values are applied before the initial condition.

> > I see parameter table parameter values as changing the model definition before any simulation is initialized. This way:
> > - initial conditions $x_0$ are computed with the parameter table values, if $x_0$ is specified as initialAssignmentRules in SBML. Parameter table parameters are not timecourse parameters, so they are not similar to SBML events.
>
> I'd say this depends on how you define $x_0$. If $x_0 = x(t_0, \theta)$, then this would be the post-event state. If you don't want to consider the condition table for computing $x_0$, this means that there can't be a condition-specific $x_0(\theta)$. I don't think that's what we want.

I agree with this. As specifying the initial condition is not directly in the PEtab scope (only experiment-specific changes to the initial condition), we should support users specifying the initial condition in their model -- therefore changes from PEtab parameters and conditions should be applied before the initial condition, such that the initial condition can be dependent on them.

> > - in gradient-based optimization, the sensitivity of the initial condition to the estimated parameter $\frac{\partial x_0}{\partial \theta}$ will be computed with the estimated parameter value, as opposed to the model value. If the initial condition is defined in the model by initialAssignmentRules, as opposed to the PEtab condition table, the parameter table values will never be used for $x_0$ or $\frac{\partial x_0}{\partial \theta}$ -- but they should be.
>
> Hm, in gradient-based optimization, we usually don't really care about $\frac{\partial x_0}{\partial \theta}$, but only about $\frac{\partial \mathcal{L}}{\partial \theta}$, so this wouldn't really be an argument for or against either, would it?

It would be problematic but to clarify, I see now that I was describing the situation that you already described above: does the PEtab user need to write in their manuscript that: "The parameter was estimated, except for when the model's initial condition was evaluated, where it took the value defined in the model." ?

For this reason, I don't agree with the "SBML event" perspective for the parameter table values and experiment condition changes. I think the "pre-initial condition" perspective is more intuitive for users and has no unexpected consequences that would require warnings, except for those in PEtab v1 already. Treating condition changes as SBML events with highest priority is fine for the intermediate timepoints, but not for the initial timepoint due to the model's initial condition.

dweindl:
> I think the "pre-initial condition" perspective is more intuitive for users 

Agreed.

> and has no unexpected consequences that would require warnings, except for those in PEtab v1 already.

Partially agreed. The unexpected consequences might be the potentially very different results obtained from applying some condition at $t_0$ vs. $t_0 + \epsilon$, but I think that's something one can quickly get used to, and I am not sure how relevant that is in practice.

dweindl:
I think we have a couple of interconnected issues here. I'll try to summarize my understanding of the different concerns and potential solutions brought up so far. 

Main problem: The current specs (`main`) do not specify where exactly the parameter table comes into play.

We want to address that while achieving (most of) the following objectives:

**Objective 1**: The parameter table values must be usable when evaluating targetValues of the condition table. Otherwise, it won't be possible to estimate any initial conditions in a experiment-specific manner. 

**Objective 2**: Changing some parameter globally through the parameter table *should* have the same effect as changing it for every experiment through the condition table, at least for the cases where we set a parameter to some constant value (which is the only thing supported in the parameter table anyways). That was the initial purpose of the condition table in PEtab v1, and I think that was good. Now in PEtab v2, this is a bit more complex, because we don't just define initial conditions, but also changes *during* a simulation and we allow for more complex expressions.

**Objective 3**: Ideally, condition table changes have the same effect independent of when they are applied during an experiment.

**Objective 4**: Compartment size changes should preserve the amounts of contained species. (Not so essential for me personally, but I think this was agreed upon earlier. I don't think it was ever clarified whether this should also hold for parameter table changes.)

**Objective 5**: Be able to make use of initial assignments contained in the model while using PEtab condition specific overrides to avoid completely moving them to the condition table.

I think these objectives are mutually exclusive (happy to be proven wrong) and the weighting will probably be very subjective. For me, currently, Objective 1 >> 2 >> 5 ~= 3 > 4.

What can we do?

**Option 1**: One way to achieve Objective 1 would be introducing an additional initialization step after model initialization but before experiment-initialization (first initialize the model without any PEtab changes, then apply the parameter table changes, then apply the condition table changes).
This would also cover Objective 3&4 and mostly cover Objective 2, except for cases when there are other targetValue expressions that depend on the changed value, but it is incompatible with Objective 5.

**Option 2a**: Another way to achieve Objective 1 would be applying parameter table changes before model initialization. This way, the parameter table changes would mirror SBML initial assignment. However, this conflicts with Objective 2, because condition table changes are currently interpreted as event assignments. Changing a compartment size through an initial assignment may have a effect than changing it through an event. It covers Objective 3, though. It covers Objective 4 for the condition table, and covers Objective 5 for the parameter table.

To resolve the issue in Option 2a, we have at least the following options:

**Option 2b**: Apply the parameter table changes before model initialization and always treat the first period of each experiment as initial assignments instead of event assignments. This generally makes sense to me, but it means that a given condition has different implications depending on when it is applied during a timecourse (first vs. subsequent period), and thus, interferes with Objective 3. That's the cost of conflating initial assignments and event assignments in the condition table. It does not cover Objective 4 for the condition table for the first period of an experiment, but it covers Objective 5.

**Option 2c**: Apply the parameter table changes before model initialization and treat all condition table changes as initial assignments. This covers Objectives 1-3,5, but is incompatible with Objective 4.

For all of the above options, the condition table changes could be converted to SBML events for simulation, they would just look different.

Looking forward to hear other opinions / solutions.

dilpath:
Thanks for clarifying these aspects. I think Option 2c is the best option, for the objectives you described as well as some additional reasons:
- Option 2c is backwards-compatible with PEtab v1, because in v1 we have Objectives 1, 2, 3, and 5
- Option 2c doesn't involve SBML semantics (e.g. SBML event semantics). This means that a PySB user doesn't need to read the SBML spec in order to understand PEtab experiments
- Option 2c is probably easier to understand even for SBML users, since SBML event semantics are complicated in some cases

doc/v2/documentation_data_format.rst
```diff
@@ -1103,6 +1102,52 @@ Conditions and observables that are not applied to a model do not need to be
...
+The following describes how the model is initialized and how other PEtab
+parameter values are applied before the simulation starts.
+This procedure is executed for each model and experiment combination
+defined in the PEtab problem.
+
+1. Pre-initialization
+
+   1. Parameters values for parameters that occur in the parameter table are
+      applied to the uninitialized model.
+      For estimated parameters, the respective externally provided values
```
dilpath, line 1119:
```suggestion
      For estimated parameters, the respective externally-provided values
```

doc/v2/documentation_data_format.rst
```diff
@@ -1103,6 +1102,52 @@ Conditions and observables that are not applied to a model do not need to be
...
+      are applied and for non-estimated parameters, the nominal values
+      from the parameter table are applied.
+
+   2. The time is set to the start time of the first period of the given
+      experiment (or 0 if there is no explicit experiment), and any
+      experiment-specific conditions provided for the first period are applied
+      (see :ref:`v2_reinitialization_semantics` for details).
+      Target value expressions for the first period must not refer to model
+      entities that are not also listed in the parameter table.
```
dilpath, line 1128: Possible to also allow model entities listed in the first period condition changes? This would mean that the order of application of changes now matters.

dilpath (approved):

doc/v2/documentation_data_format.rst
dweindl: In principle, yes, as long as there aren't any circular dependencies. However, this would be inconsistent with later periods, where the respective symbols evaluate to their value at the end of the previous condition, not to the newly assigned value.

dweindl:
> I will shortly update the pull request accordingly.

Done. Please let me know if this is in line with your expectations.

doc/v2/documentation_data_format.rst
```diff
@@ -1097,6 +1096,52 @@ Conditions and observables that are not applied to a model do not need to be
...
+
+The following describes how the model is initialized and how other PEtab
+parameter values are applied before the simulation starts.
+This procedure is executed for each model and experiment combination
+defined in the PEtab problem.
+
+1. Pre-initialization
+
+   1. Parameters values for parameters that occur in the parameter table are
+      applied to the uninitialized model.
```
FFroehlich, line 1112: Clarify what uninitialized means here?

FFroehlich (approved):

doc/v2/documentation_data_format.rst
dweindl: This will probably be model-format-dependent. I can add "pre-initial-assignment-execution" as an example for SBML models. Keeping SBML specifics out of the PEtab specs didn't work too well anyways... :see_no_evil:

dweindl:
@matthiaskoenig @fbergmann @sebapersson Does this work for you?

sebapersson (approved):
Looks good to me, thanks for fixing!

matthiaskoenig (approved):
Reads good to me, have to test how easy this is to implement with roadrunner (especially setting the parameters before applying the initial assignments).

PR #646, opened 2025-07-28
Add `id` field

dweindl:
I think it would often be convenient to have some ID for a given PEtab problem and I would like to add such a field to the yaml schema. Previously, we usually used the model ID in such situations. However, often we have multiple problems based on the same model but different datasets, and this doesn't work well in the case of multiple models.

Not sure whether it should be mandatory or optional. I think, I'd go for optional.

### DIFF
--- a/doc/v2/_static/petab_schema_v2.yaml
+++ b/doc/v2/_static/petab_schema_v2.yaml
@@ -14,6 +14,11 @@ properties:
 
     description: Version of the PEtab format
 
+  id:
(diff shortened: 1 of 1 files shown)


fbergmann (approved):
I'm fine with an `id` field. I'd prefer to have it optional.

dilpath (approved):
Fine for me as either optional or required. Some documentation for it should be added to the spec.

PR #651, opened 2025-10-20
Reword observableParameters / noiseParameters description

dweindl:
Reword observableParameters / noiseParameters description and make them more self-contained. Previously, this was too specific to normal noise models.

### DIFF
--- a/doc/v2/documentation_data_format.rst
+++ b/doc/v2/documentation_data_format.rst
@@ -560,8 +560,8 @@ Detailed field description
   values. This may be used to account for condition-specific or
   batch-specific parameters.
 
-  If none of the observables use any observable placeholders,
-  this column may be omitted.
+  If none of the observables referenced in a given measurement table use any
+  noise placeholders, this column may be omitted there.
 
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):

doc/v2/documentation_data_format.rst
```diff
@@ -540,39 +540,60 @@ Detailed field description
...
+  :ref:`parameter table <v2_parameter_table>`.
+
+  Different lines for the same ``observableId`` may specify different
+  values. This may be used to account for condition-specific or
+  batch-specific parameters.
 
-  Numeric values or parameter names are allowed. Same rules apply as for
-  ``observableParameters`` in the previous point.
+  If none of the observables use any noise placeholders,
+  this column may be omitted.
```
dilpath: Fine as is but since measurements can be split across multiple tables, and not all tables need to have the same columns, then the column may be omitted in some tables if those tables do not contain relevant observables.

dilpath (approved):

PR #654, opened 2025-12-02
Add discrete noise distributions

dweindl:
Adds Poisson, binomial, and negative binomial distributions to the list of noise distributions.

Closes #653.

PR #656, opened 2025-12-17
Clarify interpretation of noise distributions

sebapersson:
When implementing support for `LogLaplace` in PEtab.jl, I realized that the interpretation of the noise distributions in the spec is not entirely clear. In particular, for the supported distributions, the model output is not assumed to be the mean or location of the data distribution, but rather its **median**.

For example, let ($m$) be the measured value, $y := \text{observableFormula}$ the simulated value, and $\sigma$ the noise. For the `LogNormal` distribution in PEtab we have $\log(m) \sim \mathcal{N}(\log(y), \sigma)$, which implies $m \sim \mathcal{LN}(\log(y), \sigma)$. For this `LogNormal`, the median is `y` (`exp` of first argument).  A similar interpretation holds for `LogLaplace`. Overall, this PR aims to clarify this.

### DIFF
--- a/doc/v2/documentation_data_format.rst
+++ b/doc/v2/documentation_data_format.rst
@@ -921,6 +921,20 @@ Detailed field description
 Prior distributions
 ~~~~~~~~~~~~~~~~~~~
 
+The prior distributions supported for the
+:ref:`MAP objective function and for Bayesian inference <v2_objective_function>`
+are listed below. PEtab only supports univariate prior distributions.
+The probability density functions (PDFs) below assume that the parameter bounds
+are wide enough to not truncate the distributions. If the parameter bounds
+are narrower than the distribution's support, the distributions are truncated,
+resulting in the following truncated prior distribution:
+
+.. math::
+   \pi_{\text{trunc}}(x) = \frac{\pi(x)}{\text{CDF}(\text{upperBound}) - \text{CDF}(\text{lowerBound})}
+
+where :math:`\pi(x)` is the PDF of the non-truncated distribution
+and :math:`\text{CDF}(\cdot)` its cumulative distribution function.
+
 Let :math:`x` denote the parameter value and :math:`\Gamma` the
 `Gamma function <https://en.wikipedia.org/wiki/Gamma_function>`__.,
 then the following prior distributions are supported:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,16 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
-    -   id: check-yaml
-    -   id: check-added-large-files
+      - id: trailing-whitespace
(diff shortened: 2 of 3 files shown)


matthiaskoenig (approved):

doc/v2/documentation_data_format.rst
```diff
@@ -748,7 +748,7 @@ Noise distributions
 
 Denote by :math:`m` the measured value,
 :math:`y:=\text{observableFormula}` the simulated value
-(the location parameter of the noise distribution),
+(the median of the noise distribution),
 and :math:`\sigma` the scale parameter of the noise distribution
```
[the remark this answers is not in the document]
sebapersson: Actually, for normal distributions it is the standard distribution (no log), while for Laplace it is the scale. I will clarify this, thanks!

doc/v2/documentation_data_format.rst
```diff
@@ -748,7 +748,7 @@ Noise distributions
 
 Denote by :math:`m` the measured value,
 :math:`y:=\text{observableFormula}` the simulated value
-(the location parameter of the noise distribution),
+(the median of the noise distribution),
```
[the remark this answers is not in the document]
sebapersson: > "first parameter in canonical notation"

This is not correct either, as with log-normal we have `log(m)` not `m` entering the formula, so I would like to keep median.  As for #654, I think the way to handle it would simply be to split the noise distribution into two sections, discrete and continuous.

doc/v2/documentation_data_format.rst
```diff
@@ -761,25 +760,31 @@ Then we have the following effective noise distributions:
   * - Type
     - ``noiseDistribution``
     - Probability density function (PDF)
-  * - Gaussian distribution
+  * - | Gaussian distribution
+      | (i.e., :math:`m` is normally distributed as :math:`m \sim \mathcal{N}(y, \sigma^2)`)
```
dilpath:
```suggestion
      | (i.e., :math:`m \sim \mathcal{N}(y, \sigma^2)`)
```

doc/v2/documentation_data_format.rst
```diff
@@ -761,25 +760,31 @@ Then we have the following effective noise distributions:
...
     - ``normal``
     - .. math::
          \pi(m|y,\sigma) = \frac{1}{\sqrt{2\pi}\sigma}\exp\left(-\frac{(m-y)^2}{2\sigma^2}\right)
   * - | Log-normal distribution
-      | (i.e., :math:`\log(m)` is normally distributed)
+      | (i.e., :math:`\log(m)` is normally distributed as :math:`\log(m) \sim \mathcal{N}(\log(y), \sigma^2)`)
```
dilpath:
```suggestion
      | (i.e., :math:`\log(m) \sim \mathcal{N}(\log(y), \sigma^2)`)
```

doc/v2/documentation_data_format.rst
```diff
@@ -761,25 +760,31 @@ Then we have the following effective noise distributions:
...
     - ``log-normal``
     - .. math::
          \pi(m|y,\sigma) = \frac{1}{\sqrt{2\pi}\sigma m}\exp\left(-\frac{(\log m - \log y)^2}{2\sigma^2}\right)
-  * - Laplace distribution
+  * - | Laplace distribution
+      | (i.e., :math:`m` is Laplace distributed as :math:`m \sim \mathcal{L}(y, \sigma)`)
```
dilpath:
```suggestion
      | (i.e., :math:`m \sim \mathrm{Laplace}(y, \sigma)`)
```

PR #663, opened 2026-01-26
Remove SBML local parameter mapping

dweindl:
As discussed during a recent PEtab editor meeting, we don't want to support SBML local parameters via the mapping table, as this just introduces unnecessary complexity. Instead of aliasing local parameters via the PEtab mapping table, they should just be implemented as global parameters in the SBML model to begin with.


matthiaskoenig (commented):
This looks good. We could perhaps add a sentence/clarification in the spec along the lines of: 
> SBML *local* parameters are not supported. Local parameters should be promoted to global parameters (e.g. using libsbml).

This could be as part of the parameter description.

matthiaskoenig (approved):

dilpath (approved):

PR #664, opened 2026-01-26
Update PEtab extensions

dweindl:
Rephrase and rearrange some extension-related parts:

* Trying to make the PEtab 2.0 specs more self-contained
* Separate process and specs for PEtab extensions
* Avoid using "extension" to refer to different things


matthiaskoenig (approved):
This looks good. Do we have any extensions which are currently used? If yes it makes sense to list them on the webpage.

dweindl:
> Do we have any extensions which are currently used? If yes it makes sense to list them on the webpage.

There's this: https://github.com/PEtab-dev/PEtab/blob/26c4fb251e93efe06afdb2f2500a3403bda765a1/README.md?plain=1#L112-L128

doc/v2/documentation_data_format.rst
```diff
@@ -91,10 +91,10 @@ The following sections outline the minimum requirements of those components
 in the core standard, which should provide all necessary information for
 defining a parameter estimation problem.
 
-Extensions of this format (e.g., additional columns in the measurement table)
-are allowed and encouraged. While such extensions can enhance plotting,
+Additional non-standard columns are allowed in all PEtab tables.
+While such extra column may, for example, contain metadata to enhance plotting,
```
dilpath:
```suggestion
While such extra columns may, for example, contain metadata to enhance plotting,
```

doc/v2/documentation_data_format.rst
```diff
@@ -91,10 +91,10 @@ The following sections outline the minimum requirements of those components
...
 downstream analysis, or improve efficiency in parameter estimation, they
-should not alter the definition of the estimation problem itself.
+must not alter the definition of the estimation problem itself.
```
dilpath: Does this contradict the ability of extensions to alter the definition through new columns?

Or does "non-standard" include both "core" and "extension" columns?

dilpath (approved):

doc/v2/documentation_data_format.rst
dweindl: Hm, yeah, let's add something like "except for extensions with `required: true`."

doc/development.rst
```diff
@@ -233,6 +212,12 @@ PEtab extension, it needs to go through the following process.
 #. The technical specification and documentation of the extension is submitted
    as a pull request in this repository that references the respective issue.
 
+Additionally, the following requirements need to be met:
+
+* There is at least one tool that supports the proposed extension.
```
fbergmann, line 217: is one tool really sufficient? If we are concerned about exchange, wouldn't  we need two tools? For comparison here is what SBML mandates: 

```
There exist at least two separate software implementations. There must be at least two software systems that implement support for manipulating the constructs defined in the package specification.

* The software must be able to manipulate the constructs in the package, and not simply store and retrieve the package constructs. (For this reason, implementing support for the package in a library such as [libSBML](https://sbml.org/documents/process/level-3-process/Software/libSBML) alone does not qualify as one of the implementations, although of course other software may use the libSBML package support as part of their qualifying implementations.)
* Each software package must be able to manipulate a majority of the features defined in the package. (The precise definitions of what constitutes a majority of the features and able to manipulate are left to the PWG to decide. The operational criteria may be different for different SBML Level 3 packages.)
* Every feature in the package specification must be supported and manipulated by at least one of the two tools.
```

dweindl: This what was previously written, I just moved that line.
Valid concern, though. I am open to changing that. However, I'd prefer addressing that separately, potentially together with some other policy updates.

PR #669, opened 2026-05-13
Proofreading

m-philipps:
- Clarified the mapping table
- Minor fixes to links, spelling or improving readability (subjectively)
- Suggested an alternative version of Daniel's Figure that is read from left to right instead of clock-wise. Feel free to reject.


doc/v2/documentation_data_format.rst
```diff
@@ -1069,16 +1069,10 @@ Detailed field description
 
 - ``petabEntityId`` [PETAB_ID, REQUIRED]
 
-  A valid PEtab identifier (see :ref:`v2_identifiers`) that is not defined in
-  any other part of the PEtab problem.
+  A valid PEtab identifier (see :ref:`v2_identifiers`) that is globally unique.
```
m-philipps, line 1072: This was replaced with "globally unique", because the documentation does not define "define" and the original description would make me think that the `petabEntityId` can not be a `parameterId`.

doc/v2/documentation_data_format.rst
```diff
@@ -1069,16 +1069,10 @@ Detailed field description
...
   This identifier may be referenced in condition, measurement, parameter and
   observable tables, but cannot be referenced in the model itself.
 
-  The ``petabEntityId`` may be the same as the ``modelEntityId``, but it must
```
m-philipps, line 1077: Removed because under this restriction, the ids can never be the same. Also, with allowing the `modelEntityId` column to be empty, the use case of using the mapping table for further, optional columns is already covered.

doc/v2/gfx/petab_files.svg
dweindl, line 1: I prefer the old clockwise ordering, but okay to change it if the majority prefers left-to-right.
Maybe put that into a separate PR.

doc/v2/documentation_data_format.rst
dweindl: > because the documentation does not define "define"

I see that point, but I don't think this change resolves it. If it isn't used in any place other than the mapping table, the `petabEntityId` is rather useless. If it is used somewhere else, one still needs to distinguish between definition and reference.

> the original description would make me think that the `petabEntityId` can not be a `parameterId`.

So far, this was the assumption because we didn't consider any model format where model parameters could have names that aren't compatible with PEtab IDs.
To support that case, the parameter table description would have to be updated as well.
This would be a format change then and should go to a different PR.

doc/v2/documentation_data_format.rst
dweindl: > the ids can never be the same

But that's explicitly allowed here, isn't it? I don't consider an identity mapping an alias.

The main point here was to prevent recursive mappings such as `A => B; B => C; C => D; ...` to keep the things simpler. I think this should be preserved. If the goal is changing that, please move it to a dedicated PR.

dweindl (commented):
Thanks, the typesetting changes look good to me. Please move the suggested PEtab changes to dedicated PRs for further discussion.

doc/v2/documentation_data_format.rst
dilpath: Yes, perhaps we can rewrite this part
> The petabEntityId may be the same as the modelEntityId, but it must not be used to alias an entity that already has a valid PEtab identifier. This restriction is to avoid unnecessary complexity in the PEtab problem files.

to be like
> The mapping table must not be used to alias an entity that already has a valid PEtab ID. However, the special case of `petabEntityId == modelEntityId` is not considered an alias and can be useful for annotations.

